PR #158, opened 2015-07-13
Adding missing README from svn to fix make install breakage

amchoukir:
Hello,

During `make install` the target `install_doc` complains about README missing, I looked around the issues and did not find something reported for this. The way I fixed it for now is by copying the README from the original svn repo: `https://distcc.googlecode.com/svn/trunk/README`. Let me know if that is ok with you?

Cheers

Amine

### DIFF
--- a/README
+++ b/README
@@ -1,0 +1,55 @@
+
+	  distcc -- a free distributed C/C++ compiler system
+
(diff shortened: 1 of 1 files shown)


thedrow:
Thanks!

PR #161, opened 2015-09-25
Localslots with gcc -E

afbjorklund:
Currently the --localslots are only read when distributing,
not when running locally.  Make sure to always parse the hosts
configuration, for localslots. We want to use --localslots_cpp
for running gcc -E, so use a special exit code to signal this.
- This program is free software; you can redistribute it and/or
- modify it under the terms of the GNU General Public License
- as published by the Free Software Foundation; either version 2
- of the License, or (at your option) any later version.
  *
- This program is distributed in the hope that it will be useful,
- but WITHOUT ANY WARRANTY; without even the implied warranty of
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU General Public License for more details.

### DIFF
--- a/src/where.c
+++ b/src/where.c
@@ -83,6 +83,22 @@ static int dcc_lock_one(struct dcc_hostdef *hostlist,
                         int *cpu_lock_fd);
 
 
+void dcc_read_localslots_configuration()
+{
+    struct dcc_hostdef *hostlist;
+    int ret;
+    int n_hosts;
+
+    if ((ret = dcc_get_hostlist(&hostlist, &n_hosts)) == 0) {
+        while (hostlist) {
+            struct dcc_hostdef *l = hostlist;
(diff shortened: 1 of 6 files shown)


src/where.c
```diff
@@ -83,6 +83,22 @@ static int dcc_lock_one(struct dcc_hostdef *hostlist,
                         int *cpu_lock_fd);
 
 
+void dcc_read_localslots_configuration()
```
shawnl: Is the purpose of this to check for syntax errors in the host list? Because the code doesn't seem to do anything.

shawnl: oh, nevermind

shawnl: ahh I see in dcc_parse_hosts()

shawnl:
oh damn I didn't mean to squash the commits

PR #162, opened 2015-10-16
BackslachInMacro: resolve make check failure

tones111:
This change resolves a failure when running the BackslashInMacro test due to the following compiler warning:

stderr:
testtmp.c: In function 'main':
testtmp.c:13:5: warning: implicit declaration of function 'puts' [-Wimplicit-function-declaration]

This was found on a system running gcc 5.2.0

### DIFF
--- a/test/testdistcc.py
+++ b/test/testdistcc.py
@@ -211,13 +211,13 @@ def _IsMachO(filename):
             # 0xfeedface' are also mach-o.
             contents.startswith('\xFF\xED\xFA\xCE') or
             contents.startswith('\xCE\xFA\xED\xFF'))
-    
+
 def _IsPE(filename):
     '''Given a filename, determine if it's a Microsoft PE object file or
     executable.  The magic number used ('MZ') is taken from
     /usr/share/file/magic on an ubuntu machine.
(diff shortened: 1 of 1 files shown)


paranormal:
has problem gone? why did you close it?

TafThorne:
The associated change for this ticket seems to be mostly whitespace.  The only non-whitespace alteration I could see was to move #include <stdio.h> out of a `#if FALSE` block.  I think that `puts` is defined in that library so that may fix it.  

```
@@ -966,8 +966,8 @@ def source(self):
  class BackslashInMacro_Case(ComputedInclude_Case):
      def source(self):
          return """
 +#include <stdio.h>
  #if FALSE
 -  #include <stdio.h>
    #define HEADER MAKE_HEADER(testhdr)
    #define MAKE_HEADER(header_name) STRINGIZE(foobar\)
    #define STRINGIZE(x) STRINGIZE2(x)
```

paranormal:
I see, with -Wall it'll get a warning. But it'll link to standard library and find its puts eventually.

paranormal:
nevertheless, why not.

PR #164, opened 2015-11-04
Add help for systemd-based RH-like distro users

vincele:
Recent RedHat-like distributions run distcc through systemd and
the configuration is not taken from this file, so add a comment that
put the poor user in the right direction.

This patch is RFC here because the downstream bug has not been
fixed and is going to be EOL'ed

See: https://bugzilla.redhat.com/show_bug.cgi?id=1187517

If this is not the way it should be fixed^Wworkarounded, please advise...


thedrow:
Thanks!

PR #167, opened 2015-12-28
Handle 64-bit time_t on OpenBSD

jasperla:
On OpenBSD, time_t is defined as a 64-bit type. So adjust the format string accordingly and make sure this gets casted correctly for other systems too.

### DIFF
--- a/src/lsdistcc.c
+++ b/src/lsdistcc.c
@@ -479,8 +479,8 @@ void server_handle_event(state_t *sp)
 
         if (opt_verbose > 2)
             fprintf(stderr,
-                    "now %ld %ld: server_handle_event: %s: state %d\n",
-                    now.tv_sec, (long) now.tv_usec/1000,
+                    "now %lld %ld: server_handle_event: %s: state %d\n",
+                    (long long) now.tv_sec, (long) now.tv_usec/1000,
                     sp->req.hname, sp->status);
 
         switch (sp->status) {
@@ -506,8 +506,9 @@ void server_handle_event(state_t *sp)
                 /* start the nonblocking connect... */
                 if (opt_verbose > 0)
                     fprintf(stderr,
-                            "now %ld %ld: Connecting to %s\n",
-                            now.tv_sec, (long) now.tv_usec/1000, sp->req.hname);
+                            "now %lld %ld: Connecting to %s\n",
+                            (long long) now.tv_sec, (long) now.tv_usec/1000,
+                            sp->req.hname);
                 if (connect(sp->fd, (struct sockaddr *)&sa, sizeof(sa))
                     && errno != EINPROGRESS) {
                     if (opt_verbose > 0)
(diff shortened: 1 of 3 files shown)


paranormal:
Yeap. But it would fit in %ld, no need to cast it to (long long).

paranormal:
I mean the current code would work with 64 bit time as well. It does for me (FreeBSD) at least. Please put down your compiler warnings, etc.

jasperla:
Hi,

I'm getting errors such as:

```
src/compile.c: In function 'dcc_build_somewhere_timed':
src/compile.c:821: warning: format '%ld' expects type 'long int', but argument 4 has type 'time_t'
```

`time_t` is defined as `__time_t` which is defined as `__int64_t` which in turn is a `long long`.

paranormal:
I see. My fault. Will look soon.

jasperla:
cheers!

paranormal:
Could you cast it to long instead of long long and watch if it fixes the warning please?

paranormal:
I believe that long and long long both 8 bites on your system, am I right?

paranormal:
Oh, sorry, it's already long. Do you use i386 and the long on your system is 32 bits?

paranormal:
We are probably good, fixing cast to long where it missed. On 64 it'll work with 64 bit time, and 32 have time to perish up to 2032 year.

jasperla:
On OpenBSD time_t is 64-bit across all platforms, also 32-bit platforms. So i386 will be just fine after 2038.

paranormal:
ok, could you please fix your patch then?
here https://github.com/distcc/distcc/pull/167/files#diff-3dc78d5c2e3afa7dcd2963e0adc5e9ddL823
add one more cast for long to delta.tv_sec and see how building goes.

paranormal:
there may be other places.

paranormal:
anyway, you haven't answered, what size do long and long long have on your machine, do you use i386?

paranormal:
Please collaborate, I have no resources to install OpenBSD to investigate why and how they implemented time_t so it's incompatible with Linux/FreeBSD realisation.

I understand that casting time_t is only way to make it cross compatible, but casting it to long long probably too much.

jasperla:
Casting to something other than `long long` isn't going to work for OpenBSD.

i386:
```
long: 4
long long: 8
time_t: 8
```

amd64:
```
long: 8
long long: 8
time_t: 8
```

paranormal:
It is, till 2038. But compiler do it anyway now, throwing a warning.

jasperla:
FWIW, AWS has an OpenBSD AMI and there are also vagrant boxes available where you can install OpenBSD onto. Perhaps that would be easier for you?

src/lsdistcc.c
```diff
@@ -743,10 +745,11 @@ static int one_poll_loop(struct rslave_s* rs, struct state_s states[],
...
             if (opt_verbose > 0)
                 fprintf(stderr,
-                        "now %ld %ld: Resending %s because "
-                        "deadline was %ld %ld\n",
-                        now.tv_sec, (long) now.tv_usec/1000, sp->req.hname,
-                        sp->deadline.tv_sec, (long) sp->deadline.tv_usec/1000);
+                        "now %lld %ld: Resending %s because "
+                        "deadline was %lld %ld\n",
+                        (long long) now.tv_sec, (long) now.tv_usec/1000,
+                        sp->req.hname, sp->deadline.tv_sec,
```
shawnl: missed a cast to (long long) here

jasperla: Actually not, since `tv_usec` is defined as a `suseconds_t` which is a `long` so it's correct as you've merged it. Thanks for that!

PR #168, opened 2016-01-13
Avoid locking the CPU for preprocessed input

afbjorklund:
It is not needed to (possibly having to first wait for and then) take a
lock on localslots_cpp for preprocessed input (such as emitted by ccache).

We are not going to start the preprocessor, and will not try to preprocess
remotely with pump. So the preprocessor lock taken will be unnecessary anyway.
- This program is free software; you can redistribute it and/or
- modify it under the terms of the GNU General Public License
- as published by the Free Software Foundation; either version 2
- of the License, or (at your option) any later version.
  *
- This program is distributed in the hope that it will be useful,
- but WITHOUT ANY WARRANTY; without even the implied warranty of
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU General Public License for more details.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -560,10 +560,12 @@ dcc_build_somewhere(char *argv[],
         goto run_local;
     }
 
-    /* Lock the local CPU, since we're going to be doing preprocessing
-     * or include scanning. */
-    if ((ret = dcc_lock_local_cpp(&local_cpu_lock_fd)) != 0) {
-        goto fallback;
+    if (!dcc_is_preprocessed(input_fname)) {
+        /* Lock the local CPU, since we're going to be doing preprocessing
+         * or include scanning. */
+        if ((ret = dcc_lock_local_cpp(&local_cpu_lock_fd)) != 0) {
+            goto fallback;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


src/compile.c
```diff
@@ -560,10 +560,12 @@ dcc_build_somewhere(char *argv[],
...
     }
 
-    /* Lock the local CPU, since we're going to be doing preprocessing
-     * or include scanning. */
-    if ((ret = dcc_lock_local_cpp(&local_cpu_lock_fd)) != 0) {
-        goto fallback;
+    if (!dcc_is_preprocessed(input_fname)) {
+        /* Lock the local CPU, since we're going to be doing preprocessing
+         * or include scanning. */
+        if ((ret = dcc_lock_local_cpp(&local_cpu_lock_fd)) != 0) {
```
shawnl: `dcc_unlock(local_cpu_lock_fd);`
below this will fail when using pump mode, spewing unnecessary errors. That unlock needs to be conditioned on `local_cpu_lock_fd != -1` like the others at 
`unlock_and_clean_up:`

shawnl:
I merged this with my change.

PR #171, opened 2016-01-21
added build hints for FreeBSD

krichter722:


paranormal:
It can be solved with just autoconf. And instead of legalising unused-parameter it's better to fix C code.
So, all those things should be fixed in different way and this ticked should be removed :1234: :1234:  = :79(15)(25):

paranormal:
For FreeBSD hints would be: portinstall devel/distcc or pkg install distcc

paranormal:
Oh, I can close tickets. I probably should then, till someone figured out that I have that privilege.

paranormal:
JK :)

Look at /usr/ports/devel/distcc they have done all job. Including some patches which would be unnecessary soon.

PR #172, opened 2016-01-21
fixed broken build on FreeBSD 10.1 due to unused parameters

krichter722:


thedrow:
We'd rather see them removed from everywhere or not at all for now.
The build should be fine for FreeBSD. We're building with -Wno-errors.

PR #174, opened 2016-03-22
distcc/test python(3) ready

paranormal:
I hope that this would be effort worthy.

Some things need just slight clean up. Some are broken by default. I'd like to move for the other things, and I think that fixing test should be the first priority. They are not in good shape and clearness, but hey, what is? I'd like to know maintainers opinion about clearing some old mess first, then may be more tight integration of pump mode. But first things first:

1)
```diff
-                "--pid-file %s --port %d --allow 127.0.0.1"
+                "--pid-file %s --port %d --allow 127.0.0.1 --listen 127.0.0.1"
```
without this it is useless for me (I'm on FreeBSD), it does weird stuff, runs itself on tcp6 instead of v4. I didn't dig much though.
2) Compile_c_Case wouldn't work because tests check mtime, atime and distcc code ctime. Fixing C code like this
```diff
-                if (stat_dep.st_ctime >= reference_time) {
+                if (stat_dep.st_mtime >= reference_time) {
```
would make it work. I think that C must be fixed rather then tests, but I could miss something, and it was made like this intentionally...

### DIFF
--- a/test/comfychair.py
+++ b/test/comfychair.py
@@ -105,13 +105,14 @@ class TestCase:
         """
         while self._cleanups:
             try:
-                apply(self._cleanups.pop())
+                clean_task = self._cleanups.pop()
+                clean_task()
             except KeyboardInterrupt:
-                print "interrupted during cleanups"
+                print("interrupted during cleanups")
                 _report_error(self, debugger)
                 return 2
             except:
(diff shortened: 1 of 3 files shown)


paranormal:
pushed.

Issue #155, opened 2015-06-06
distcc tcp mode is a security risk [CVE 2004-2687]


sourcefrog:
http://www.cvedetails.com/cve/2004-2687

distcc currently has two modes by which clients can connect - over TCP (default) or SSH. 

In TCP mode, distcc checks the client IP address against a whitelist, which (iirc) is required but can be set quite loosely. There is of course no guarantee that every user on a permitted client address is friendly.

Once the connection is established the client can reasonably easily manipulate the server into running arbitrary commands. This is automated in eg http://www.rapid7.com/db/modules/exploit/unix/misc/distcc_exec

Possible solutions include:
- Deprecating TCP mode and requiring SSH connections. This would largely push the problem onto SSH and make it more reasonable for distccd to trust the client.  However SSH does have a noticeable performance overhead, or to be more precise it did when I measured 10 years ago. It may have changed. It may cause problems with server discovery.
- Use some other transport that provides authentication/integrity/encryption but is faster than SSH.
- Add a simple authentication protocol. This may be cheaper than SSH. There is a risk the authentication protocol itself would be buggy. If there's only authentication of thie client without integrity checks or encryption, it may remain vulnerable to malicious networks.
- Using platform-specific security features such as `seccomp_bpf` to restrict what can be done once the compiler command is launched. This might be a useful adjunct to other measures but is probably not enough, and may be brittle.

Mathiasdm:
I would argue that, within a corporate network, it may be preferable to go with the TCP mode (trusting the local machines and users) to gain the best performance.

stevenhoneyman:
Or private isolated (i.e. trusted) network. I have no need for any security in distcc - if someone manages to get close enough to touch that, they already have complete physical access anyway!

Optional extra, great. Mandatory... not so much.

onionjake:
You can also use SSH multiplexing to only pay the overhead of establishing the ssh connection once: http://www.linuxjournal.com/content/speed-multiple-ssh-connections-same-server

The server will limit how many simultaneous sessions per network connection, so keep that in mind: http://unix.stackexchange.com/a/22987

uzytkownik:
> Add a simple authentication protocol. This may be cheaper than SSH. There is a risk the authentication protocol itself would be buggy. If there's only authentication of thie client without integrity checks or encryption, it may remain vulnerable to malicious networks.

Or just use TLS + SASL. That is essentially solved problem and a lot of effort is done to make it both fast and secure for browsers.

[...]

Issue #134, opened 2015-06-06
missing feature: support --coverage to compile remotely


GoogleCodeExporter:
```
It would be great if distcc would support --coverage.

So distcc needs to send the generated gcno file back in addition to the obj 
file. (short version)

```

Original issue reported on code.google.com by `climbera...@gmail.com` on 2 Jan 2014 at 10:59

PR #177, opened 2016-04-14
include_server python(3) ready

paranormal:
Any review, testing, approval, or disapproval would be appreciated.


paranormal:
Testing by building subversion 4 times with bdb, (without external dependency the building time would be the same).

without pump:
      322.52 real       105.53 user        67.21 sys
      345.07 real       104.81 user        68.94 sys
      354.99 real       104.65 user        67.90 sys
      341.88 real       104.84 user        69.56 sys

with pump:
      267.56 real       107.61 user        67.57 sys
      270.66 real       107.19 user        68.05 sys
      271.03 real       106.81 user        68.44 sys
      270.97 real       106.56 user        68.93 sys

thedrow:
LGTM

itensionanders:
I don't think raising the requirement from python2.4 to python3.1 was a reasonable change to apply. :(
Changing to python2.7, or preferrable python2.6 so that it still works on RHEL 6, would've been better...

`fatal error: Python.h: No such file or directory`
`configure: error: Python interpreter is too old`

This just increases the need for a pump opt-out ?

paranormal:
RHEL6 has old ruby, old python, old everything, it's meant to be in that way because they price stability.
Using old distcc is reasonable way then for RHEL6.

paranormal:
Anyway we discussed the possibility to support python both 2.7 and 3 at the beginning, and it's still can be done, I even have C module code working for both versions. But some more work should be done, and code would be more branched everywhere.

We concluded to make it work for 3 version of python because it's reasonable (at least for me :)). World moving towards it.

Anyway2 distcc doesn't require any modules, just install 3'd python and you are ok.
Anyway3 with this patch https://github.com/distcc/distcc/pull/178 it would be possible to build distcc without pump and python.
It's hasn't merged yet, because no one reviewed it yet and discussed it.

itensionanders:
TBH, I would probably have preferred leaving it at python2.4* even (dropping python 2.2 is fine).
But I don't think that it would have worked with keeping both python2/python3 compatibility alive...

* then it would support RHEL 5 out-of-the-box too, you see... :-)
* even the "new" RHEL 7 has python2.7 as default (not python3)

itensionanders:
Something like this: 86cd5d089f6026bd8668039a4d6d203b4dd4d9ab

paranormal:
Leaving it at python2.4* even (dropping python 2.2 is fine), -- then it would work only on RHEL. There some inconsistency in syntax between 2.7 and everything less.

Who is going to support all possible Python versions? This effort could be applied to make something more productive.

itensionanders:
The suggestion is to leave it at that for distcc-3.2, and then bump the requirement to python2.7 for 3.3 ?
Then it should be reasonable effort, to have it working with python2 (2.7 *only*) and python3 for new dev

i.e. so that you have a master branch where you do the new and exciting development (for future 3.3),
and a maintenance branch where you give the "3.2rc1" a proper release or two with vital bug fixes ?

paranormal:
I'm not the release master. Let him decide. The requested feature is that:
let it work with python 2.7 too.

itensionanders:
Yep, the required feature is that when you decide to upgrade to python3 - that you still support python2...
And that you keep "good ole python" (i.e. python2.4-python2.6-python2.7) on a 3.2 maintenance branch.

This is so that we can upgrade from distcc 3.1, eventually. And then the distcc development can continue unhindered on `master`, while 3.2 goes in a slower pace (but still gets some releases once in a while)

itensionanders:
@thedrow: are you acting as release manager ? would it be OK to keep python3 to distcc 3.3 and later ?

itensionanders:
And yes, we _would_ be willing to help out with maintaining (and testing) such a "legacy" distcc branch.

paranormal:
If there would be decision to support 2.7 - I have C code working for both 3 and 2.7, so it'll save a time to someone who would do this. You can support any python you want even 1.1 - but without me.

itensionanders:
@paranormal : are you OK with keeping the python2 support (that you have already done?), if you can use future imports and import six or whatever ? Or would that not be possible/feasible, you think ?

itensionanders:
(as in: https://docs.python.org/3/howto/pyporting.html and https://pythonhosted.org/six/ respectively)

thedrow:
All new distributions that I know of have Python 3 available or use Python 3 by default.
Do you have a list of the distributions that don't?
We're not going to release this as a patch version obviously. We can postpone the release of this feature if needed but ultimately, Python 3 is the future and distcc is not a library so we don't need to support both versions of Python if Python 3 is already available.

itensionanders:
My concern was not that python3 was _unavailable_, it can probably be compiled from source if required. It was just that it raised the "entry level", and for a feature that I wasn't particularly interested in. But if it is possible to "opt out" like suggested, then that could be a workaround if you want to go all in on python3 ?

RHEL/CentOS 5: `python2.4`
RHEL/CentOS 6: `python2.6`
RHEL/CentOS 7: `python2.7`

What do you think about a maintenance branch and releases (`3.2-maint`), and keeping the future/development for a future release (`master`) - once there are some new features ? As in: call the new version with python3 (and so on) distcc-3.3 instead of the suggested "3.2" (3.2.0 ?)

paranormal:
@itensionanders I said that I've done python import module for both version of python (distcc_pump_c_extensions_module.c), not the whole thing. Doing this for python code would be ugly, and more important pointless.

All OS's you mentioned must have a way to install python3 otherwise they wouldn't be  able to run current software.

paranormal:
The future is here. Python 3 was a future in 2008.

itensionanders:
That is fine, just don't live in the future (yet): http://www.enterpriseitadoption.com/

But I would take that as a "no", then ? You *don't* want to support python2/python3 in parallell. Which is fine, as long as it still builds without it (currently you a "Python.h" not found type of of error, when building). And I would still prefer if you called the new version 3.3, since it will be a while before we leave 3.2 behind...

paranormal:
No from me. But the whole decision is not me alone, and I'm not the only coder here.

paranormal:
I remember how RHEL had 1.6 ruby, which would core dump literally after any unusual structure or even prolong work.
You have to understand that you can install Slackware 10 year from now and complain in the same way.

thedrow:
We can keep releasing 3.2 as long as someone volunteers to do the work to make it happen.
So let's turn this into actionable items:
1) Create a new branch before this feature was merged.
2) Find a maintainer for distcc3.2 (unless @itensionanders volunteers)
3) Create a release cycle that would involve all stakeholders including both maintainers and packagers such as Ubuntu/Debian, CentOS/REHL etc.

itensionanders:
@thedrow: we would be happy to support 3.2 with development and testing, but prefer if you or someone else did the actual release ? Could also be nice if there were more people to review the suggestd patches

Let's move the discussion elsewhere (list?), and consider this item (python3) as closed then.

TafThorne:
@afbjorklund I do not mind helping out with the tagging and code reviews for actual releases.  I am not a big user of distcc but I can code C, C++ & Ruby to a reasonable standard.  If I stare at Python for a while it kind of makes sense.  

@thedrow 's suggestion of actionable items seems sensible.  I can mangle git well enough to posthumously fork a branch from not-the-HEAD of another one.  I shall generate a 3.2-maint branch from before the move to Python3 (I do not think that is what @paranormal was trying to achieve in the 3.2 branch we currently have, if it is then that is one less thing for me to do)

> Let's move the discussion elsewhere (list?), and consider this item (python3) as closed then.

That sounds reasonable too.  The Project Planning thread I started before stumbling back into here on Git already does much of that.  I will copy the proposed actions over onto the end of my thread and mention this pull request as the source.

TafThorne:
My proposed message has been sent to the list: https://lists.samba.org/archive/distcc/2016q4/004449.html 

The message I sent before coming here was: https://lists.samba.org/archive/distcc/2016q4/004448.html

I started the Project Planning thread with: https://lists.samba.org/archive/distcc/2016q4/004446.html to which /Anders posted a reply.  


In the latest message I said:

> Dropping Python 2 Support
> 
> (or more specifically moving up to python2.7 from python2.4 on a
> maintenance branch and dropping python3 support from master)
> 
> Having looked into starting a 3.2 release or 3.2_maintenance branch I
> stumbled into the https://github.com/distcc/distcc/pull/177 request. 
> Under that request the creation of the 3.2 maintenance branch is
> suggested and it is proposed that said branch will be used to maintain
> some support for Python2.7 for as long as it seems necessary.  
> 
> The main work would continue on master, with the creation of a 3.3.#
> release in the near future.  The master branch (and therefore the 3.3
> releases onwards) would require python3. 
> 
> 
> What do the viewers of this list think of the proposal? 
> 
> 
> 
> Silence will be taken as support.  If the action is supported I will try
> an action at least some of it at the end of the month.

That seems like enough multi-cross-posting.  As the emails suggest I will try and action some items at the end of the month if no one objects on either platform.

paranormal:
@afbjorklund why wouldn't you write support for 2.7 simultaneously with what have already being done? I mean it was there, it wouldn't be as hard as writing support for new version. Everyone would be happy if the code would work with both versions and may be then you would have some appreciation for other peoples work, especially if the work is not one hour hacking.

TafThorne:
Supporting two versions is always going to be more work than supporting one.  That being said it would be a shame to drop support on widely used versions (which I believe 2.4 and 2.7 are).  

The above comment by @paranormal is a little confusing to me though as @afbjorklund was arguing for the retention of 2.7 after paranormal removed it as far as I could see? I have probably lost the thread somewhere.  

I have a feeling that this thread is dragging out much longer than it needs to.  I am not a Python guy, if this was a Ruby 1.7 vs. Ruby 2.1 thing I could weigh in but it is not.  If someone wants to explain to me that Python 3 is like Ruby 1.9 through 2.1 while 2.7 is like 1.7 then I might at least get the analogy. Anyway to the point.  

If someone has a patch that will keep Python 2.4 through 3.x working for this project today that seems a reasonable thing to apply.  If we do not _need_ to drop support for older version today for this fix then I think keeping it is of value.  We could then differ the switch till later.  If that patch does not exist and supporting older versions would be burdensome then let us go ahead with the creation of a python2.7 (and lower?) maintenance branch as suggested under https://github.com/distcc/distcc/pull/178.  To do that we need to work out a sensible point to merge from to create the branch and put https://github.com/distcc/distcc/pull/182 on.

paranormal:
I'm confident in Ruby more then in Python. Unlike Ruby, Python third-party developers stick to 2.7 and don't like moving forward, it frustrates language developers but somehow it's how it works. Ruby community is lucky for having what they have now...

2.4-2.6 is out of question. Current standard is 2.7.

And yes it took longer that it should, In my view ideal solution would be having both versions, 2.7 and 3.x. A little worst is backing to 2.7.

afbjorklund:
The request was to _delay_ the requirement of either python2 (i.e. 2.7) or python3 until a maintenance branch had been forked off. I don't mind that any new development is done using either, and it's OK if it is python3-only (maintaining both with e.g. "six" is somewhat of a hassle, and maybe best avoided)

The reason was that we were using distcc 3.1/3.2 with ccache 3.1/3.2, and weren't using the "pump" functionality anyway. Now with distcc 3.3 (master) and ccache 3.3, it is no longer sending preprocessed source code (but instead full source code) so I will need to look into the pump mode again for that...

And for that python3 is fine. If it *has* to be Python still, maybe it could just be written in C instead ?
(I guess if someone forked of an incompatible Ruby 3000, they too would be reluctant to change.)

afbjorklund:
As for new features, beyond suggested bug fixes, the most interesting would be `--coverage` (#134)

And I suppose that [SASL](https://en.wikipedia.org/wiki/Simple_Authentication_and_Security_Layer) _could_ be added as a feature, to keep any anxious types satisfied ? (#155)

Arguing about Python versions for distcc is not fun...

And maintenance branches are by definition "dead".

So I think that we should just stick with python3 on master. And do something useful with it instead ?

Choosing python2 for maint wouldn't help much, then one would still have to install something "extra".

PR #178, opened 2016-05-02
3.2 to go

paranormal:
A brief overview.
-  3.2 is a suggested version
-  --docdir is already covered by autoconf and we don't have to redefine it.
- --disable-pump-mode a good way to disable include_server at all (in case you don't have the Python, etc)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,11 +9,10 @@ dnl Process this file with autoconf to produce a configure script
 # Copyright (C) 2002, 2003, 2004 by Martin Pool
 
 # As of 0.6cvs, distcc no longer uses automake, only autoconf.
-AC_PREREQ(2.53)
-AC_INIT(distcc, 3.2rc1, distcc@lists.samba.org)
+AC_PREREQ(2.69)
+AC_INIT(distcc, 3.2, distcc@lists.samba.org)
 
 AC_CONFIG_HEADERS(src/config.h)
-AC_CANONICAL_HOST
 
 # FreeBSD installs its version of libpopt into /usr/local/, but does
 # not put that on the default library and header path.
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -9,11 +9,10 @@ dnl Process this file with autoconf to produce a configure script
 # Copyright (C) 2002, 2003, 2004 by Martin Pool
 
 # As of 0.6cvs, distcc no longer uses automake, only autoconf.
-AC_PREREQ(2.53)
-AC_INIT(distcc, 3.2rc1, distcc@lists.samba.org)
+AC_PREREQ(2.69)
+AC_INIT(distcc, 3.2, distcc@lists.samba.org)
 
 AC_CONFIG_HEADERS(src/config.h)
-AC_CANONICAL_HOST
```
paranormal: leftover from something

itensionanders:
I think disabling pump mode is a reasonable feature, but then it should probably #ifdef the code too ?

paranormal:
Sorry don't understand, where exactly do you want to ifdef it?

paranormal:
Currently the server would be able to serve pump mode (by C code) even with --disable-pump and without any python.
Client wouldn't have pump mode part build which is python.

There are a flaw in --disable-pump and what it's really does. But there also divided functionality for pump mode, server C, client Python. And I don't know what is better, more branching? --without-python and --without-pump? Or just like it is now. It wouldn't require the Python and it would serve for pump as a additional feature.

itensionanders:
Where `DCC_CPP_ON_SERVER` is being used, i.e. only for pump mode

And requiring python (for pump) is OK, it is python3 that is the issue here.

itensionanders:
Currently we are using distcc with ccache, so not using pump mode but sending precompiled headers.
But it _could_ be used in the future, by using the `CCACHE_CPP2` feature (i.e. running the cpp twice).

itensionanders:
Anyway, I would rather see a `3.2-maint` branch created where you would keep "python" (python2.4-python2.7) as the version being used, and maybe add some of our bug fixes to it instead ? ;-)

Then you could work with raising the python version requirements on the `master` branch, but it would still be preferred if *both* "python2" (i.e. python2.7) and "python3" could still be used there too...

itensionanders:
The bugs that I mentioned would be #160 and #161

itensionanders:
Created pull request #182 (for 3.2), the python3 change could be re-applied on master after branching out.
(still, if you had already had a version that worked with *both* of python2 and python3 that would be great!)

But I would prefer if any upcoming 3.2 did **not** have that change.
Also the feature to disable pump mode seems a bit premature ?

afbjorklund:
New version was released as `3.3` instead, with `python3` (thanks) and all the other new stuff:

https://github.com/distcc/distcc/pull/266

This means that we could still do a `3.2-maint` (with python2.4) if there is enough interest in it...

So this PR can probably be closed now ?

PR #180, opened 2016-05-25
Add UNUSED macro to make CC happy with -Werror option.

uframer:
GCC complains about unused arguments in popt library if -Werror is on (which is the default configuration). Add a UNUSED macro in popt.h and apply it on every unused argument.


martinm82:
I wouldn't make modifications in the popt library but instead disable them with a compiler flag (see as well https://github.com/distcc/distcc/pull/226/commits/ee99c3953e7e9f48636e7083251e31b5f0d74383)

shawnl:
This is a smörgåsbord of changes.

PR #182, opened 2016-05-25
Keep requirements at Python 2.4 (for now), instead of 3.1

afbjorklund:
As discussed in #177 (after it was merged, unfortunately - didn't see the python discussion on the list)...

It would also be OK to keep python3 on master, and apply this change on a new `3.2-maint` branch ?


TafThorne:
> It would also be OK to keep python3 on master, and apply this change on a new 3.2-maint branch ?

Yes that would seem sensible... if I can remember how to do it.

shawnl:
@rmatev the problem I see is that doesn't getting rid of UnicodeDecodeError while using python 3 require making the code python3-only? If that is the case releasing for python 2 means we really release ONLY for python 2.

rmatev:
@shawnl it is possible to make the code compatible with both python2 and 3, see http://python-future.org/compatible_idioms.html#file-io-with-open

shawnl:
Then I would love to see 3.2 release with python 2 support. I want to release in the next couple weeks. Using python 2 by default seems like a good idea for 3.2.

afbjorklund:
Seems like master was released (though confusingly still versioned as 3.2), requiring `python3`.
The suggesion was to keep `python` for 3.2 maintenance, but I suppose that will be "hard" now.

@shawnl : would probably have been a good idea to bump the new release to "3.3" or something ?
Since it now has a lot more changes, than was detailed in NEWS. Including the change of python.

shawnl:
I could re-release it as 3.3 (and in the process note the changes in NEWS).
I just knew that we needed to release. Only then can distros pick it up.

afbjorklund:
@rmatev : while it is "doable" to keep support for `python2` (this means python2.7 and no earlier versions) and `python3`, the author of the python upgrade wasn't interested (see #177) in doing it or supporting it.

afbjorklund:
@shawnl : if you could do that, I think it would be good.

Hopefully there will be more releases anyway, like 3.3.1

afbjorklund:
Thanks, then I guess we now say that 3.3 requires `python3` and 3.2 still works with `python`.

PR #183, opened 2016-06-17
FIX: NULL is undeclared without stddef.h so the MAP_FAILED test fails

apeabody:
The "if mmap() supports MAP_FAILED" test in configure.ac didn't include <stddef.h> leaving NULL undeclared on most modern (at least the ones I tested) C compilers, and therefore the test always failing.


shawnl:
LGTM

PR #186, opened 2016-06-17
FIX: disable unused-parameter warning for bundled popt

### LINKED ISSUE #185: popt version
Hi Everyone,

A quick update that I wasn't able to compile with the bundled version of popt under CentOS-7.  However using the bundled CentOS-7 version of popt-devel 1.13 worked like a champ.  I might try and drop the most recent version (http://rpm5.org/files/popt/) in the tree as I don't know how common popt is outside RHEL and Debian given it's RPM heritage.

Cheers,

Andrew

apeabody:
This change enables the included popt to be built on "newer" gcc versions such as included in CentOS-7.

Closes #185

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -209,8 +209,8 @@ then
     PYTHON_CFLAGS="$PYTHON_CFLAGS -Wno-missing-prototypes -Wno-missing-declarations \
 -Wno-write-strings"
 
-    # For popt/*.c, we disable unused variable warnings.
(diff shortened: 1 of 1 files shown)


martinm82:
Can this change be merged to master? I have that as well fixed in one of my PRs but this one is lingering since some time already.

shawnl:
This was commited in 6dadc5f64b6

Issue #145, opened 2015-06-06
zeroconf hosts have a hardcoded slots value of 4?


GoogleCodeExporter:
```
I noticed that the value for the number of jobs on a zeroconf host is hardcoded 
to 4 in hosts.c

if ((ret = dcc_zeroconf_add_hosts(ret_list, ret_nhosts, 4, ret_prev) != 0))


This has been causing all kinds of issues for me when testing it - for some 
reason, with 1 zeroconf host, and localhost/4... i get a `distcc -j` value of 
20! If this is still being developed/maintained; would it be possible to send 
the number of available job slots when using zeroconf/avahi to detect them?

Unless I'm missing something obvious... which is possible :)
```

Original issue reported on code.google.com by `stevenho...@gmail.com` on 9 Sep 2014 at 9:09

ericfont:
This explains a lot.   This has been giving me problems when trying to get an even distribution of work across my slaves.  One slave is 3 cores, for which distcc --show-hosts reports 12 slots, while another slave is 8 cores, for which distcc --show-hosts reports 32 slots.  Unfortunately, my machine with lower core count just gets overwhelmed.  I could never figure out quite where those numbers came from or how to change them until I found this issue.

Anyway, I think user should be able to specify number of jobs for each slave by using the . "-j" argument when use "--zeroconf".  Should also probably stick to the same default of NUM_CORES + 2 which I believe is used elsewhere.

ericfont:
The following line sets arg_max_jobs (the number of permitted jobs) from the -j "jobs" parameter:

https://github.com/distcc/distcc/blob/b2fa4e21b4029e13e2c33f7b03ca43346f2cecb8/src/dopt.c#L124

Then these lines set dcc_max_kids to arg_max_jobs or if -j wasn't specified, then fallsback to one job per CPU + 2:

https://github.com/distcc/distcc/blob/b2fa4e21b4029e13e2c33f7b03ca43346f2cecb8/src/dparent.c#L118-L121

But then right after that in the HAVE_AVAHI block, the code currently does zeroconf registration using 
"n_cpus":

https://github.com/distcc/distcc/blob/b2fa4e21b4029e13e2c33f7b03ca43346f2cecb8/src/dparent.c#L141-L147

[...]

PR #187, opened 2016-07-03
fix Issue #145. Allow -j (jobs) param to set number slots when used in zeroconf mode.

ericfont:
If server is started in zeroconf mode, the previous behavior was that the number of slots for any server was hardcoded to 4 * number of cpus.  This made zeroconf impracticle when different hosts had different core counts.
This commit allows distccd to be invoked in zeroconf mode with the -j parameter to specify total slots for the particular server being invoked.
If no -j specified, then this host will publish its total number of slots to be 2 * number of cores + 2.  (Note: this is the same behavior as when invoking a non-zerconf server without specifying -j.)
To maintain backward compatibily, if a distccd zeroconf server doesn't publish a value for slots, then it will be assigned a default total slots equal to the previous behavior of 4 * number of cores.
Note that the meaning of "slots" in the code now always refers to the number of slots per host.  (The previous interepration of "slots" was inconsistent, because sometimes it meant the number of slots per core.)

### DIFF
--- a/man/distccd.1
+++ b/man/distccd.1
@@ -215,6 +215,8 @@ DNS Service Discovery (DNS-SD).  This allows distcc clients on the local
 network to access this distccd server without explicitly listing its host
 name or IP address in their distcc host list: the distcc clients can
 just use "+zeroconf" in their distcc host lists.
+Can optionally use -j parameter to specify the maximum number of jobs
+that this server can process concurrently.
 .B This option is only available if distccd was compiled with
 .B Avahi support enabled.
 .TP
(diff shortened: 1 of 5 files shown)


ericfont:
I forced updated this PR's commit so that I'm using the word "jobs" to refer to the maximum number of jobs per host, instead of using the word "slots".

TafThorne:
Thank you.  

As this looks to be a helpful extension of existing functionality I am happy to merge in the request.  No one else has issued any objections in the last week either.

PR #188, opened 2016-07-15
Portability fixes for distcc

jimklimov:
These pedantic fixes and configure-variants extensions were required for compilation on OpenIndiana Hipster. Verified that build succeeds on Debian 8 with these in place.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -19,6 +19,11 @@ QUICK SUMMARY
    Note that past versions of Debian used binutils-dev instead of
    libiberty-dev.
 
+   Note that distcc-pump needs Python 3.x, while the default version of
+   Python on many systems is 2.x, so you may have to be explicit - e.g.
+
+       sudo apt-get install libpython3.5-dev
+
 1. Build and install
 
       ./autogen.sh       # If "configure" does not already exist.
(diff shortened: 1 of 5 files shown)


jimklimov:
https://github.com/OpenIndiana/oi-userland/pull/2167

For idea of casting to `(struct sockaddr *)`  thanks to Alexander Pyhalov (@pyhalov).

timofonic:
Any news?

afbjorklund:
:+1: LGTM

But the comment about python is not _entirely_ true, since the test suite was *also* ported to python3.
So it's not only distcc-pump that requires it now, and you will need both python3 and libpython3-dev.

Assuming that Python 3.6 would work too, and not only Python 3.5, but haven't really verified that...

jimklimov:
well... not all devs build distcc itself (and with a full testsuite on top), so they can care about the pump but not about the test requirements  ;)

jimklimov:
Building without pump sort of defeats the idea of providing a fix to build with one ;) But still, thanks for the easy workaround to users of common master branch so far ;)

jimklimov:
Cool, thanks ;)

Issue #157, opened 2015-06-06
migrate docs from googlecode


sourcefrog:
distcc documentation is currently visible at http://distcc.googlecode.com/svn/trunk/doc/web/index.html

This will go away in 2016 so we should republish them on github.

n1tehawk:
A "quick and dirty" way to do this (without having to fall back on github pages) is http://rawgit.com/. E.g. you could point users to https://cdn.rawgit.com/distcc/distcc/master/doc/web/index.html - links still pointing to _code.google.com_ would have to be adjusted, of course.

sourcefrog:
Nice idea ,thanks.

On Wed, Dec 9, 2015 at 2:01 PM, NiteHawk <notifications@github.com> wrote:

> A "quick and dirty" way to do this (without having to fall back on github
> pages) is http://rawgit.com/. E.g. you could point users to
> https://cdn.rawgit.com/distcc/distcc/master/doc/web/index.html - links
> still pointing to *code.google.com <http://code.google.com>* would have
> to be adjusted, of course.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/distcc/distcc/issues/157#issuecomment-163408692>.
>



-- 
Martin

TafThorne:
I have implemented the suggestion by @n1tehawk and referenced our 3.1 release branch code.  rawgit.com asks that you refere to tags in published pages so they can cache results.  

If this is good enough for now we could mark this issue closed.  I do not mind spending time later trying to migrate to github pages if people would prefer that.  With the current implementation we will need to update the link used so that we do not accidentally get stuck on the 3.1 version of the documents forever.

PR #190, opened 2016-07-26
Add tracis-ci.org Build To distcc Project

TafThorne:
These changes add a .travis.yml file to the root directory of the project.  With this in place it would be possible for an administrator to start a travis-ci.org build for the project to cover a Debian/Ubuntu native build.  The service is free for Open Source projects. 

As I am not very good at using GitHub there is also ee86ccf on the end.  This updates the web documentation link as suggested by @n1tehawk under #157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 before_install: 
  - sudo apt-get update -qq
- - sudo apt-get install -qq make binutils-dev gcc-dev
+ - sudo apt-get install -qq make binutils-dev
 
 language: cpp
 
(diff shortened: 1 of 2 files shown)


TafThorne:
The travis-ci.org build for my branch can be seen at: https://travis-ci.org/TafThorne/distcc 

This shows the very simple build process in action.  An administrator of the main distcc project could enable a similar setup for the main project.

TafThorne:
@thedrow could you please consider activating the travis-ci.org build for now.  I know we talked via email about using https://circleci.com/ instead as it might allow us to actually use distcc in the build process.  For now I would rather get some basic automated Linux builds working, then add OS X and self referencing builds later.

thedrow:
I just enabled distcc on Travis CI. Sorry it took so long.

TafThorne:
Thank you for enabling that.  It will make me a little more confident to approve changes once we can see them at least compiling on several systems.

PR #191, opened 2016-08-11
ssh.c: tokenize the ssh command retrieved from DISTCC_SSH

prwolfe:
Basically we have a large and ever-changing user community
to support and cannot rely on corporate to put ssh options
into the machine configurations.  This allows us to place
-o ErrorLevel=ERROR in the DISTCC varable and have distcc
properly interpret that for passing on to execvp.

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -193,8 +193,10 @@ int dcc_ssh_connect(char *ssh_cmd,
                     pid_t *ssh_pid)
 {
     pid_t ret;
+    const int max_ssh_args = 12;
+    char *ssh_args[max_ssh_args];
+    char *child_argv[10+max_ssh_args];
     int i,j;
-    char *ssh_args[6];
     int num_ssh_args = 0;
 
     /* We need to cast away constness.  I promise the strings in the argv[]
@@ -207,6 +209,8 @@ int dcc_ssh_connect(char *ssh_cmd,
         while (token != NULL) {
             ssh_args[num_ssh_args++] = token;
             token = strtok(NULL, " ");
+            if (num_ssh_args == max_ssh_args)
+                break;
         }
     }
     if (!ssh_cmd)
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -178,6 +178,12 @@ static int dcc_run_piped_cmd(char **argv,
  * nice for us to parse it into an argv[] string by splitting on
  * wildcards/quotes, but at the moment this seems redundant.  It can be done
  * adequately using .ssh/config I think.
+ *
+ * @note the ssh command does need to be tokenized as we have hundreds of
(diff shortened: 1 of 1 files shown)


src/ssh.c
```diff
@@ -187,14 +193,22 @@ int dcc_ssh_connect(char *ssh_cmd,
...
+    char *ssh_args[6];
+    int num_ssh_args = 0;
 
     /* We need to cast away constness.  I promise the strings in the argv[]
      * will not be modified. */
 
-    if (!ssh_cmd)
-        ssh_cmd = getenv("DISTCC_SSH");
+    if (!ssh_cmd) {
+        char *ssh_cmd_in = getenv("DISTCC_SSH");
```
TafThorne: Please could all of the variables be declared at the start of the function, as there were previously.

src/ssh.c
```diff
@@ -187,14 +193,22 @@ int dcc_ssh_connect(char *ssh_cmd,
...
+        ssh_cmd = strtok(ssh_cmd_in, " ");
+        char *token = strtok(NULL, " ");
+        while (token != NULL) {
+            ssh_args[num_ssh_args++] = token;
```
TafThorne: Is there anything that protects against num_ssh_args becoming more than the 6 allocated in line 197?  Could we use a `MAX_SSH_ARGS = 6` or a `const int` and have the while loop exit early if the limit is reached?

src/ssh.c
```diff
@@ -205,8 +219,13 @@ int dcc_ssh_connect(char *ssh_cmd,
     if (!path)
         path = (char *) "distccd";
 
+    char *child_argv[10+num_ssh_args];
```
TafThorne: 10 looks like a magic number, possibly relating to the value found in: 
`char *child_argv[10];` 
However this was replaced with `char *ssh_args[6];` in your patch.  
Also could the variable declaration be moved to the start of the function please?

TafThorne:
I have made some in line notes about the patch.  Please let me know what you think of my suggestions.  The goal of the patch seems worthwhile to me.

src/ssh.c
prwolfe: Yeah - I debated about this.  We don't really know that this is the limit and we should dynamically determine the size at run-time.

src/ssh.c
prwolfe: I moved it so we could allocate the correct size (although I should have used the heap for this if I really wanted to do that.)  I would only favor moving it back if we set a larger max and check our bounds as you requested in the note at line 208.  I can do that.

prwolfe:
Somehow the repository got a duplicate commit.  I am closing this and will re-submit with the corrections TafThorne and I discussed.

PR #192, opened 2016-08-12
ssh.c: tokenize the ssh command retrieved from DISTCC_SSH

prwolfe:
We have a large and ever-changing user community
to support and cannot rely on corporate to put ssh options
into the machine configurations. This allows us to place
-o ErrorLevel=ERROR in the DISTCC varable and have distcc
properly interpret that for passing on to execvp. Includes
changes from previous review by TafThorne

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -178,6 +178,12 @@ static int dcc_run_piped_cmd(char **argv,
  * nice for us to parse it into an argv[] string by splitting on
  * wildcards/quotes, but at the moment this seems redundant.  It can be done
  * adequately using .ssh/config I think.
+ *
+ * @note the ssh command does need to be tokenized as we have hundreds of
+ * users and a corporate requirement that keeps us from modifying the
+ * system ssh config files. We can at the same time set command-line options
+ * through the tool in use one level above this. - prw 08/09/2016
+ *
  **/
 int dcc_ssh_connect(char *ssh_cmd,
                     char *user,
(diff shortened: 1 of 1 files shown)


src/ssh.c
```diff
@@ -187,14 +193,26 @@ int dcc_ssh_connect(char *ssh_cmd,
...
-    if (!ssh_cmd)
-        ssh_cmd = getenv("DISTCC_SSH");
+    if (!ssh_cmd) {
+        char *ssh_cmd_in = getenv("DISTCC_SSH");
+        ssh_cmd = strtok(ssh_cmd_in, " ");
+        char *token = strtok(NULL, " ");
+        while (token != NULL) {
+            ssh_args[num_ssh_args++] = token;
+            token = strtok(NULL, " ");
+            if (num_ssh_args == max_ssh_args)
```
TafThorne: Should this be `if (num_ssh_args >= max_ssh_args)`?

src/ssh.c
```diff
@@ -207,6 +225,10 @@ int dcc_ssh_connect(char *ssh_cmd,
 
     i = 0;
     child_argv[i++] = ssh_cmd;
+    for (j=0; j<num_ssh_args; ) {
+        child_argv[i++] = ssh_args[j++];
```
TafThorne: Would it not be a more common behaviour to perform the increments to `i` and `j` in the `for` braces rather than inline?  That would also let you use the more effect, although these days not really any more effect in most optimised compilers, pre-increment rather than post.  You could probably more the i = 0 inside the for loop initialisation too as it is related to it.

src/ssh.c
```diff
@@ -187,14 +193,26 @@ int dcc_ssh_connect(char *ssh_cmd,
                     pid_t *ssh_pid)
 {
     pid_t ret;
-    char *child_argv[10];
-    int i;
+    const int max_ssh_args = 12;
+    char *ssh_args[max_ssh_args];
+    char *child_argv[10+max_ssh_args];
+    int i,j;
```
TafThorne: Having got to the end of the function.  Is the `j` itterater needed?  It looks like `i` could be used in its place for the current code.

TafThorne:
Thank you for taking the comments from the previous conversation into consideration.  I have made a couple more on the new code.  None of the current points I make are very significant, so I would be happy to merge the code as is if you would rather not make any further alterations.

TafThorne:
For future notes, the previous review was performed under #191

src/ssh.c
prwolfe: not sure how it could ever get to be >, but it won't hurt to do it that way.

src/ssh.c
prwolfe: True, that's a remnant from the previous iteration.

prwolfe:
Thanks for looking it over - as usual it results in better code.  I will be out of touch for a few weeks so you can merge out as it stands or I will push a new commit when I get back and can test it. I also think the current comments are minor enough to leave as is.

TafThorne:
You are welcome.  I am pleased to try and review code and get things added in where I can.  This is a very useful tool and unless we can get a community behind it to keep up support it would lapse, as many open source tools have before.

PR #194, opened 2016-08-13
distcc: remove whitespaces from code

ramsudharsan:
-> Partially reduces size of code
-> Improves indentation of code

Signed-off-by: Ramsudharsan <ramsudharsanm@gmail.com>


TafThorne:
I do not think this adds a lot of value to the existing code base. All of the changes on line with other content will dilute the revision history.  Unless you feel confident in answering any technical questions on the lines you have modified, when they are brought to you following a `git blame` I would suggest things are left as they are.  

>Partially reduces size of code
2256 chars is not a huge reduction (assuming 1 space removed on every altered line).  You have added a large diff to the version control system so transfer is unlikely to be improved and I would expect most parsers & compilers to be quite efficient at skipping it.  

-> Improves indentation of code
None of your changed lines seem to alter the indentation of the code.  In theory consistent whitespacing can improve this but in this instance it has not been altered.  

Are we permitted to alter the whitespace of the GPL?  Perhaps you have corrected it and the extra lines were the modification.  Something to check.  

I have no objection to the removal of unnecessary whitespace on lines that are otherwise blank.  It seems a needless addition to the revision history though.  I would prefer that some clear guidelines on what should be done be added and that authors tidy up the code as they modify adjacent lines.

ramsudharsan:
These changes are an experiment of an automated script that I was working on. The indentation part is just an extension of my script which I haven't applied onto the code. I accept that it dilutes the revision history. I don't think this would help you. I am hereby closing this pull request.

TafThorne:
Thank you for being understanding with my comments.  

Maybe your script could be extended to have a `--only-blank-lines` option?

PR #196, opened 2016-09-15
typo fix

galeksandrp:


TafThorne:
Thank you for spotting that and for providing a patch to fix it.

PR #198, opened 2016-09-22
Handle arguments needed to build Swift on MacOS

dabrahams:
These arguments are used by clang but until now, unrecognized by distcc.


dabrahams:
I can explain the glob, at least: "-F" introduces a "framework path."  A framework is a bundle of related libraries (usually one) and headers under a single directory.  Adding a framework path effectivaly adds all of its */Headers directories to the -I #include path.

dabrahams:
Bump.  Please, I don't like maintaining my own fork and permission to post these fixes back to the project was hard-won.

TafThorne:
There have been no objections so I shall merge in your request.

dabrahams:
Thank you!

PR #200, opened 2016-10-05
fix compilation error without ENABLE_RFC2553

semverchenko:
src/access.c:55:22: error: «allones8» defined but not used
[-Werror=unused-const-variable=]
 static const uint8_t allones8 = 0xffU;


TafThorne:
No one objected so I have merged the change.

PR #203, opened 2016-10-24
Try a new host after failing to connect instead of immediately falling back to localhost

anthonybilinski:
Currently when the client fails to connect to a server, it immediately goes to fallback and compiles locally. This causes problems for me when there are many servers and a couple are offline, as the client ends up compiling many files instead of just passing to other available servers. With this change, once all the hosts are attempted and found to be be "disliked", fallback still happens and localhost compiles, so it should not reduce robustness.

### DIFF
--- a/include_server/parse_command.py
+++ b/include_server/parse_command.py
@@ -24,6 +24,7 @@ __author__ = "opensource@google.com (Craig Silverstein, Nils Klarlund)"
 import re
 import os
 import sys
+import glob
 
 import basics
 import cache_basics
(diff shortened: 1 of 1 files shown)


anthonybilinski:
My change to the printed warning "failed to distribute %s to %s, running locally instead" is to "... every host..." is only valid if we're in fallback because the hostlist has no more valid hosts, not if it's in fallback for any other reason. This isn't currently checked for, but keeping it as the original way makes it seem like if every host fails to connect, we're compiling locally because of only the last one, which doesn't seem totally fair. Neither option seems great, but the original "%s to %s" seems better since it still works for every other fallback case.

anthonybilinski:
updated commit at https://github.com/distcc/distcc/pull/204

PR #204, opened 2016-10-25
Try a new host after failing to connect instead of immediately falling back to localhost

anthonybilinski:
Currently when the client fails to connect to a server, it immediately goes to fallback and compiles locally. This causes problems for me when there are many servers and a couple are offline, as the client ends up compiling many files instead of just passing to other available servers. With this change, once all the hosts are attempted and found to be be "disliked", fallback still happens and localhost compiles, so it should not reduce robustness.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -72,6 +72,21 @@ int dcc_scan_includes = 0;
 static const char *const include_server_port_suffix = "/socket";
 static const char *const discrepancy_suffix = "/discrepancy_counter";
 
+static void bad_host(struct dcc_hostdef *host, int *cpu_lock_fd , int *local_cpu_lock_fd)
+{
+   if (host)
+       dcc_disliked_host(host);
+
+   if (*cpu_lock_fd != -1) {
+       dcc_unlock(*cpu_lock_fd);
+       *cpu_lock_fd = -1;
+   }
+   if (*local_cpu_lock_fd != -1) {
+       dcc_unlock(*local_cpu_lock_fd);
+       *local_cpu_lock_fd = -1;
+   }
+}
+
 static int dcc_get_max_discrepancies_before_demotion(void)
 {
     /* Warning: the default setting here should have the same value as in the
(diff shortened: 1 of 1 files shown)


afbjorklund:
@anthonybilinski : Maybe there should be a timeout involved here, instead of trying _all_ servers ?
or some other tradeoff, between the "try 1 or local" and "try everyone" (which could take a while...)

anthonybilinski:
For compiling a project, the only real difference is instead of:
try one, fail, temporarily blacklist host
compile local
try one, fail, temporarily blacklist host
compile local
no hosts left
compile local * 500

it looks like:
try one, fail, temporarily blacklist host
try one, fail, temporarily blacklist host
no hosts left
compile local * 500

so I think the existing timeout and blacklisting of hosts is sufficient to prevent any worse behaviour in the case where you have more files than servers, since in both cases you will still end up trying each server and timing out.

I did a benchmark compiling ffmpeg with 20 fake local network hosts (192.168.1.1/24) and watched distccmon-gnome which showed the behaviour I described above (both timing out all the servers, just in different order.)

with my modification:
time make -j4 CC="../../anthonybilinski-distcc/distcc gcc"
real	3m6.636s
user	8m32.544s
sys	0m19.588s

with the existing implementation:
time make -j4 CC="../../distcc/distcc gcc"
real	3m7.686s
user	8m33.876s
sys	0m19.464s

so as you can see, the user and real portions are basically identical as expected, and the "real" portion which is modified by timing out on servers is still basically identical.

In the case where you have fewer files than hosts, i.e. 5 files and 20 hosts, my modification would timeout on all 20 servers, whereas the existing behaviour would timeout on only 5, and this would save you a few seconds. This doesn't seem like a realistic use case to me.

anthonybilinski:
FWIW I've been using this patched version of distcc for over a year with a group of about 15 computers with usually a couple offline, and we collectively haven't run into problems. It speeds up our build a reasonable amount since the client can spend all of their CPU preprocessing jobs and distributing instead of splitting time between preprocessing and compiling work that could be distributed.

anthonybilinski:
@shawnl sorry to nag, but I continue to use this change with noticeable performance benefits with a group of desktops. Is there anything I could do to help get this merged?

shawnl:
Looks good, because the next host will still be subject to backoff heuristics.

PR #207, opened 2017-01-18
Add fallthrough comment to avoid GCC 7 error

davidlt:
Adding a comment should be enough to avoid GCC 7 error on fallthrough in switch statement. Alternatively one can use C++17 attribute which is available in C++11 and C++15 under `gnu::` or in general GNU attribute syntax.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -469,6 +469,7 @@ static void dcc_stats_process(struct statsdata *sd) {
         break;
     case STATS_COMPILE_OK:
         dcc_stats_update_compile_times(sd);
+        /* fallthrough */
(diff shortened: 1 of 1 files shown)


afbjorklund:
:+1: LGTM

TafThorne:
I'll take @afbjorklund 's comment as an approval.

afbjorklund:
Sure, it was! Eventually one might want to have a gcc-7 build in Travis, but it's a bit tricky to set up (mostly due to that it uses a "slightly" older Ubuntu, while GCC 7 is brand new and only in a PPA).

* https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

PR #211, opened 2017-04-14
Fix compilation with musl libc

doughdemon:
Only the first commit is strictly necessary; the other commits fix compilation with -Werror. (musl warns when <sys/poll.h> or <sys/signal.h> is used.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -254,7 +254,7 @@ AC_C_BIGENDIAN
 
 # Some of these are needed by popt (or other libraries included in the future).
 
-AC_CHECK_HEADERS([unistd.h sys/types.h sys/sendfile.h sys/signal.h])
+AC_CHECK_HEADERS([unistd.h sys/types.h sys/sendfile.h])
 AC_CHECK_HEADERS([ctype.h sys/resource.h sys/socket.h sys/select.h])
 AC_CHECK_HEADERS([netinet/in.h], [], [],
 [#if HAVE_SYS_TYPES_H
--- a/src/access.c
+++ b/src/access.c
@@ -52,7 +52,9 @@
  */
 
 static const uint32_t allones = 0xffffffffUL;
+#ifdef ENABLE_RFC2553
 static const uint8_t allones8 = 0xffU;
+#endif
 
 /**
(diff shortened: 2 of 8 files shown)


afbjorklund:
:+1: **LGTM**: shouldn't be any problem for glibc either, to apply this cleanup

afbjorklund:
Compile error is probably the same as in #200

TafThorne:
I have merged #200.  @afbjorklund do you think that means we no longer need the changes in this pull request?

afbjorklund:
@TafThorne: You probably want to add the others too, to avoid the warnings with musl

afbjorklund:
Here are the errors (warnings, really), when building with `musl-gcc`:

```
In file included from src/exec.c:61:0:
/usr/include/x86_64-linux-musl/sys/poll.h:1:2: error: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Werror=cpp]
 #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
  ^
cc1: all warnings being treated as errors
```
```
In file included from src/serve.c:71:0:
/usr/include/x86_64-linux-musl/sys/signal.h:1:2: error: #warning redirecting incorrect #include <sys/signal.h> to <signal.h> [-Werror=cpp]
 #warning redirecting incorrect #include <sys/signal.h> to <signal.h>
  ^
```
```
cc1: all warnings being treated as errors
In file included from src/lsdistcc.c:91:0:
/usr/include/x86_64-linux-musl/sys/poll.h:1:2: error: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Werror=cpp]
 #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
  ^
cc1: all warnings being treated as errors
```
```
In file included from src/rslave.c:33:0:
/usr/include/x86_64-linux-musl/sys/poll.h:1:2: error: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Werror=cpp]
 #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
  ^
cc1: all warnings being treated as errors

```

See https://www.musl-libc.org/how.html

`CC=musl-gcc ./configure --enable-Werror --without-libiberty`

PR #214, opened 2017-04-29
Fix broken "make distcheck"

afbjorklund:


TafThorne:
Broken builds are not much use.  Merging this in now.

paranormal:
Looks good. 967 - there's no need to include stdio it would be attached anyway after linking, but better with it anyway.

PR #217, opened 2017-07-06
fixed for Cygwin

cn00:
a little fix for compile and use distcc[d] in Cygwin


src/rpc.h
```diff
@@ -49,3 +52,6 @@ int dcc_r_argv(int ifd,
                const char *argc_token,
                const char *argv_token,
                /*@out@*/ char ***argv);
+
+
+#endif
```
TafThorne: A newline at the end of the file would be nice.  I am a traditionalist.

src/rpc.h
```diff
@@ -24,6 +24,9 @@
                 /* His hand is stretched out, and who shall turn it back?
                  * -- Isaiah 14:27 */
 
+#ifndef __DISTCC_RPC_H__
+#define __DISTCC_RPC_H__
+
```
TafThorne: I am surprised that this was not already in here.  Thank you for the addition.

src/exec.c
```diff
@@ -61,7 +61,9 @@
 #include <sys/poll.h>
 
 #ifdef __CYGWIN__
-    #define NOGDI
+    /* #define NOGDI */
```
TafThorne: If the define is not longer needed, I would rather it was removed than commented out.  I guess someone should check the behaviour of a handful of CYGWIN versions .

TafThorne:
Over 1200 new lines, I would hate to see what a large fix was.  

Quick look and config.sub seems to have been brought up to the latest version from:  http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub which seems fine to me.  We possibly should try and automate something to keep things more up to date or at least nag us if we we are behind.  So that is 611 lines that could be checked with a diff later.  

Similarly config.guess has been collected from http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess so that can be checked with a diff.  Another 1144 lines not to eyeball.  

The rest of the changes look OK to me.  


If someone could double check it all works OK on Cygwin that would be great.  If the majority of the changes are from an upstream source and we currently have a broken Cygwin compile, I am tempted to approve them and hope.  I'll give it a few days though incase anyone else has comments.

src/rpc.h
cn00: Sorry for this, I can't agree with you more.

src/exec.c
cn00: Sorry, I prefer remove those unused code 2 years before, but my teammates changed me.

PR #223, opened 2017-08-09
Add requirements and document defaults, add matrix

### LINKED ISSUE #222: Expand Travis CI Build Matrix
We should try to expand the Travis build matrix to cover as many cases as we can.  musl-gcc looks like an easy one to add.  Several of the recent Issues and Pull Requests have been about build issues: #207, #188, #167, #217, #214, #211.

afbjorklund:
Default platform is linux and gcc, maybe osx later.

Unfortunately can't test things like BSD / Cygwin.

Make sure to include requirements for all defaults.

Don't bother checking the extras like gnome or gtk.

Closes #222.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,30 @@ addons:
         packages:
             - python3-dev
             - libiberty-dev
+            - libavahi-client-dev
+            - musl-tools
+
+os:
+    - linux
+
+compiler:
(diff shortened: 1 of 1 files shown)


afbjorklund:
Hmm, need to make sure that python uses the system compiler for extensions...

PR #226, opened 2017-09-05
Extend Travis CI configuration for macOS/clang

martinm82:
This is an attempt to extend the Travis CI configuration for building distcc on macOS using Clang. In order to get it working I needed to adjust as well some tests to cope with clang as well.

Apart from that I have created as well some Dockerfile's to easily run builds in an isolated environment (also nice for macOS users)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.o
 *.d
+*.out
+*.err
 # binaries
 distcc
 distccd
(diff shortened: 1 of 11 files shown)


TafThorne:
I had a quick look over it all and things seem fine to me.  I am not a Python guy though so if another pair of eyes could look over it that would be great.  If not I am happy to merge it in to the main line and hope for the best.

thedrow:
This looks great! Thank you.

Issue #224, opened 2017-08-15
python3 include server fails with UnicodeDecodeError


gjkloosterman:
When using Python3, the include server fails with an internal error on encountering non-UTF-8 data:
```
WARNING include server: Preprocessing locally. Include server internal error: '<class 'UnicodeDecodeError'>: ('utf-8', b'......
```
In my case, an included header happened to contain a comment with an iso-8859-1 encoded copyright sign.

The problem disappeared after I reverted to Python2 by applying #182

martinm82:
In case you still want to make use of Python3 I applied in my fork a possible fix for that error. I could not easily switch to Python 2 so decided to fix it.

https://github.com/tt-gf/distcc/commit/87c46e379885acb4ece84e3e06bc20ccae42a24c

I know the solution is not perfect but I did not want to introduce the Python 'magic' module as otherwise the user would need to install more dependencies.

afbjorklund:
@martinm82 : can you make a pull request ? I guess _someone_ still wants `python3` on master.

martinm82:
@afbjorklund, found more unicode issues. So once I verify it with a full build I will create a PR

shawnl:
Are not all the characters we are interested in parsing ASCII, so can't we just ignore non-ASCII bytes, and pass them through as-is? Does windows-1252 or ISO-8859-1 give us that? (UTF-8 does not because of invalid sequences) that way we add no dependencies.

shawnl:
should be fixed with 23d36291

PR #227, opened 2017-09-05
Fixing Python3/unicode issues

martinm82:
This PR is an attempt to fix the Unicode decoding errors reported in https://github.com/distcc/distcc/issues/224.


include_server/test_data/supported_encoding.cpp
```diff
@@ -0,0 +1,2 @@
+�
```
TafThorne: Is this change needed?  It might just be the web interface on GitHub making it look a little strange to me.

TafThorne:
As @afbjorklund asked for the request, if he could green light the changes I am happy to do the merging.  Made my one query from a quick look over the code.  Again, not a Python guy so cannot comment much.

include_server/test_data/supported_encoding.cpp
thinrope: Probably... it is an "empty" file in ISO-8859 encoding (as reported by `file`).

martinm82: Yes, it is actually an empty file with ISO-8859 encoding used by the unit test

martinm82:
I ran also the changes against the full travis ci configuration I created in another PR in my personal branch: https://travis-ci.org/tt-gf/distcc/builds/275796576. This branch contains as well other changes but as a prove that the Python3 changes work also on the other configuration it should be enough.

afbjorklund:
@TafThorne: Actually I thought the whole python3 thing a bit premature to start with, but since it has been merged to master it would be great if it could made functional again (i.e. fix the error reported in #224).

timofonic:
Any news? Why is the distcc project stagnating while icecream gets developed faster? What about merging both projects and avoid the mess?

afbjorklund:
@timofonic : I'm not sure what the main differences/benefits between icecream and distcc are ?
The source bases look quite different though, even different languages (C/C++) and everything.

martinm82:
Basically the architecture is quite different in both projects, icecream uses a client/server approach with the icecc daemons and the schedulers that are dispatching the compile requests to the daemons. Also icecream is strictly depending on doing preprocessing locally and depending on toolchain environments (ICECC_VERSION).

Whereas distcc supports remote preprocessing and uses the locally installed build toolchains.

We evaluated both and yes, I would favour of having a mix of both projects, basically having a smart scheduler (icecc) with the ability to do remote processing (distcc) and using as well the locally installed toolchain (distcc)

thedrow:
I think we should consider to use https://pypi.python.org/pypi/chardet to detect the encoding of the file. I'm aware it is another dependency to add but it will be much more reliable.

shawnl:
Should be fixed with 23d362910b1a6b2dca8.

PR #235, opened 2017-11-04
daemon: add seccomp filter on Linux

shawnl:
This makes distcc much more secure as it also applies to child processes
(gcc and clang).

### DIFF
--- a/src/dopt.c
+++ b/src/dopt.c
@@ -115,6 +115,11 @@ enum {
 int opt_zeroconf = 0;
 #endif
 
+#ifdef HAVE_SECCOMP_FILTER
+/* Flag for enabling/disabling Zeroconf using Avahi */
+int opt_no_seccomp = 0;
+#endif
+
 const struct poptOption options[] = {
     { "allow", 'a',      POPT_ARG_STRING, 0, 'a', 0, 0 },
 #ifdef HAVE_GSSAPI
@@ -152,6 +157,9 @@ const struct poptOption options[] = {
 #ifdef HAVE_AVAHI
     { "zeroconf", 0,     POPT_ARG_NONE, &opt_zeroconf, 0, 0, 0 },
 #endif
+#ifdef HAVE_SECCOMP_FILTER
+    { "no-seccomp", 0,   POPT_ARG_NONE, &opt_no_seccomp, 0, 0, 0 },
+#endif
     { 0, 0, 0, 0, 0, 0, 0 }
 };
 
@@ -186,6 +194,9 @@ static void distccd_show_usage(void)
 "    --stats-port PORT          TCP port to listen on for statistics requests\n"
 #ifdef HAVE_AVAHI
 "    --zeroconf                 register via mDNS/DNS-SD\n"
+#endif
+#ifdef HAVE_SECCOMP_FILTER
+"    --no-seccomp               do not sandbox\n"
 #endif
 "  Debug and trace:\n"
 "    --log-level=LEVEL          set detail level for log file\n"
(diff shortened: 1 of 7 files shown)


shawnl:
can you help me add system calls I missed? the tests are passing on my machine

shawnl:
this should probably be distcc 3.3. Debian (and thus Ubuntu) is sadly shipping 3.1 so that it doesn't have ipv6 support.

shawnl:
I think this needs a run flag to turn it off, in case gcc or clang change to break it

thinrope:
@shawnl Can you elaborate a bit more on why this is good/needed?
* Why do you think it "makes distcc much more secure" ?
* How does it influence performance?

I had a quick look at [snadbox-seccomp-filter.c](https://github.com/distcc/distcc/pull/235/files#diff-ce2fdc40d5644539b6ca73f36adf5079) and I see it limits the syscalls to about 65 (of about 333 on my system):
````bash
curl -s https://raw.githubusercontent.com/shawnl/distcc/a7f7c316577627153185510fc9bac9de20e5ab09/src/sandbox-seccomp-filter.c |perl -ne 'print "$1\n" if (/SC_ALLOW\(__NR_(.*)\)/)'|sort
````
However, due to the nature of the process, things like `read`, `write`, `listen`, `unlink`, `rmdir`, etc. needs to be on the whitelist. And since we are forking possibly various compilers, the list will possibly grow.

I am definitely neither an expert in seccomp-bpf, nor malware developer, but it probably is not a hard challenge to write/adopt a piece of code to pass this coarse net. And there is a high chance that whatever hits the net will be simply ignored by endusers as just "yet another bug in distcc;  try without :-/"

Since this is a security enhancing patch, we need to evaluate what it brings (how much more secure we get) vs. what it takes (speed[1], added complexity, code maintainability/support for new compilers). 

[1] After all distcc is/was intended to speed up compiling, before anything else.

**Comment by shawnl** [comment]
I have worries that it could be fragile  too. There is another way.: if socket  activation support is added to distcc then systemd' s seccomp black lists (esp. Networking) can be used. I will have to measure the performanc  impact too.

**Comment by thinrope** [comment]
Putting `systemd`[1] in the soup (or simply adding more dependencies) usually is not a good thing.
But before anything else, to add a security patch/feature, there should be human description in how it helps, what scenarios have been considered and what needs to be considered implementation/maintenance wise.

[1] Disclaimer: I am allergic to `systemd`, LoL, but that is a personal thing.

**Comment by shawnl** [comment]
It wouldn't add any more    dependencies ( my zstd patch does that however, and given that distcc has zero dependencies otherwise makes it a big deal), it would only look at environment variable Listen_fds to see if fd 3 is already open for it.

File: src/dopt.c [hunk 1]
```diff
@@ -115,6 +115,11 @@ enum {
 int opt_zeroconf = 0;
 #endif
 
+#ifdef HAVE_SECCOMP_FILTER
+/* Flag for enabling/disabling Zeroconf using Avahi */
```
sourcefrog: comment seems wrong, at least

sourcefrog:
I mostly agree with thinrope's comment.

The PR would be better and easier to review if you said more about what the effect is meant to be, and why, rather than making reviewers infer it from the code.

We should measure performance, but I doubt it will be decisive. distcc loads are typically dominated by userspace CPU than syscalls, and seccomp inside the kernel should be relatively cheap.

The list of syscalls allowed by this seems so broad that I'm not sure it's going to stop any attacks. Or, put another way: what syscall do you think an attacker _must_ use, that is blocked by this?

I think a better sandboxing approach would be: make a container holding only the compiler and its dependencies, and run the jobs inside that.

FlorianHeigl:
Uh, so was this ever included later on?
It would be bad if this "you didn't explain well enough and so we did nothing instead" got in the way of an incomplete, but incrementally positive improvement.

Issue #225, opened 2017-08-29
-gsplit-dwarf / .dwo / DebugFission


tomazos:
When I compile with gcc and use the -gsplit-dwarf option, it produces two files, a .o and a .dwo file:

    $ cat > test.C
    void f() {}
    ^D

    $ g++ -c test.C -gsplit-dwarf

    $ ls
    test.C  test.dwo  test.o

However when I do the same thing with distcc, it only produces the .o file:

    $ distcc g++ -c test.C -gsplit-dwarf

    $ ls
    test.C  test.o

Is there anyway to configure or adapt distcc to send back both the .o file and .dwo file?

afbjorklund:
Supporting split-dwarf requires a change to the distcc protocol (i.e. to send the dwo files).

However, until then it _should_ know that it has extra files, and fall back to running locally:

...
``` C
            } else if (!strcmp(a, "-gsplit-dwarf")) {
                rs_log_info("compiler will emit .dwo files; must be local");
                return EXIT_DISTCC_FAILED;

```
...

tomazos:
Falling back to local isn't going to work for us, we need the compilation to be distributed.  I guess we need to add this feature to distcc ourselves.  Is there anyone from the distcc team that can supervise the code change if you want the feature upstream?

ktns:
Hi.

I've been also suffering from the same problem. 
I'm wondering if we can avoid the protocol change by calling a remote compiler without ```-gsplit-dwarf``` and then calling ```objdump --extract-dwarf``` and ```objdump --split-dwarf``` locally.
Wouldn't it be a possible temporary workaround?

tomazos:
It's not clear to me if (1) `--extract-dwo` and `--strip-dwo` will produce exactly the same `.o`/`.dwo` pair of files as `-gsplit-dwarf`, (2) nor is it clear if performance will be the same (if the files are split in a postprocessing step, this may produce an unnecessary intermediate use of the filesystem that `-gsplit-dwarf` could avoid).

It's definately worth testing though.  Thanks for the idea.

ktns:
[...]

PR #236, opened 2017-11-05
protocol version 2.1 (4 in wireshark): zstd compression support

shawnl:
Over crappy WiFi every bit counts. zstd give better compression than
lzo without taking much longer.

Does not overallocate decompression memory buffer like lzo.

I am trusting the benchmarks here: https://github.com/facebook/zstd
Compressor name 	Ratio 	Compression 	Decompress.
zstd 1.1.3 -1 	2.877 	430 MB/s 	1110 MB/s
zlib 1.2.8 -1 	2.743 	110 MB/s 	400 MB/s
brotli 0.5.2 -0 	2.708 	400 MB/s 	430 MB/s
quicklz 1.5.0 -1 	2.238 	550 MB/s 	710 MB/s
lzo1x 2.09 -1 	2.108 	650 MB/s 	830 MB/s
lz4 1.7.5 	2.101 	720 MB/s 	3600 MB/s
snappy 1.1.3 	2.091 	500 MB/s 	1650 MB/s
lzf 3.6 -1 	2.077 	400 MB/s 	860 MB/s


shawnl:
Just adding zstd support was a much smaller patch, but I fixed the protocol deficiency while I had the change.

shawnl:
This should probably be distcc version 4

afbjorklund:
Wouldn't you need to add this to pump protocol as well, if not wanting to keep both compressors ?

Also wondered if it wouldn't be easier to use higher LZO compression levels, and keep the protocol...

afbjorklund:
Nope, scratch that. There's only one alternative (LZO1X-999), and it is for precompressed data.

shawnl:
> Wouldn't you need to add this to pump protocol as well, if not wanting to keep both compressors ?

I think we should keep the old protocol. From lzo's web site: miniLZO consists of one C source file and three header files. It compiles to less than 6 kB (on x86), and the sources are just a few kB when packed - so there's no more excuse that your application doesn't support data compression :-) http://www.oberhumer.com/opensource/lzo/

I could add a protover 3.1, I just don't use pump mode so I didn't bother, but if that would get you to merge this I would write it.
> Also wondered if it wouldn't be easier to use higher LZO compression levels, and keep the protocol...

LZO can't compete. _Zstd_ has hight compression modes which make it more like xz, but long compression times are unacceptable for distcc's use case.

afbjorklund:
Seems like it needs to look for zstd versions, or use some compatibility stuff perhaps ?

```
src/compress-zstd.c:38:25: fatal error: zstd_errors.h: No such file or directory
compilation terminated.
```

afbjorklund:
> LZO can't compete.

Should have guessed, am using lz4 for ccache. Then again, there gzip (-1) _could_ compete...

Like you say probably want to keep both, i.e. lzo/lz4 and gzip/zstd. And use the best available ?

ccache/ccache#118, if bored

shawnl:
you system doesn't have /usr/include/zstd_errrs.h ? but you have /usr/include/zstd.h ??????

afbjorklund:
Yup, Ubuntu 16.04. (zstd 0.5.1)

shawnl:
OK try the new version that avoids errs.h

afbjorklund:
There's some syntax stuff, and that there's that casting business...
```
src/compress-zstd.c:38:3: error: invalid preprocessing directive #zstd
 # zstd hasn't always shipped zstd_errors.h
   ^
src/compress-zstd.c:38:12: error: missing terminating ' character [-Werror]
 # zstd hasn't always shipped zstd_errors.h
            ^
src/compress-zstd.c: In function ‘dcc_r_bulk_zstd’:
src/compress-zstd.c:217:24: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     } else if (out_len == -ZSTD_error_dstSize_tooSmall) {
                        ^
```
i.e. unsigned vs signed

afbjorklund:
You can probably use ZSTD_VERSION_NUMBER as well, anything over 1.1.0 should be OK.

shawnl:
OK, this time it will work with old zstd

afbjorklund:
Note that Travis is still on Ubuntu 12.04! (as in: 2012). Is there a "minizstd" ? :-)
Seriously, might need to have the support for the new protocol to be optional...

afbjorklund:
Other than, looking good! Compiles and has some promising "reviews"...
Making it optional and adding a 3.1 would be awesome, but not required.

shawnl:
>  Is there a "minizstd" ? :-)

No. I guess I can make it optional again.

afbjorklund:
@shawnl : also, want to lose those changes done to "NEWS"

shawnl:
NEWS fixed.

shawnl:
@afbjorklund are you good at python? The reason I am having difficulty making it optional is in doing conditions against HAVE_ZSTD in include_server/setup.py you can see the patch against that file to see what the problem is

afbjorklund:
You need to export the extra libraries to use, probably from the top-level Makefile (to the setup.py)

shawnl:
It is enabled by default because it is a protocol extension.

shawnl:
does https://github.com/distcc/distcc/issues/225 still need a change to the protocol? (that could be jammed in with this change?)

thedrow:
We need to fix the build to install zstd or make the zstd dependency optional.

afbjorklund:
Changing build system at the same time sounds a bit crazy, though.

thedrow:
@afbjorklund Sorry I meant the CI build script. Not the build system itself.
It looks like zstd is not whitelisted in Travis yet so we need to switch to the `sudo: required` environment or wait for https://github.com/travis-ci/apt-package-whitelist/issues/4267.

afbjorklund:
@thedrow : note that this is/was the same reason that lzo was bundled.
there are probably lots of other places/users, that *don't* have any zstd ?

thedrow:
Yes, probably. I'm not sure we need to bundle it, we just need to make it optional during ./configure.
The CI build should include it though.

afbjorklund:
@thedrow : with "buildsystem" I meant the introduction of `meson`
Can we please keep that as a separate discussion / commit  ?

Surely introduction of "zstd" can be done also with autotools ?
Preferrably it would just be some sources/headers in a sub-dir...

afbjorklund:
Builds OK here, after first installing `zstd` library.
`libzstd1-dev is already the newest version (1.3.1+dfsg-1~ubuntu0.16.04.1).`

Doing "apt install meson" only makes it worse...
`Meson version is 0.29.0 but project requires >= 0.44.`

----
So one might probably just redirect people to upstream ?
They seem to be using make to build by default, anyway.

I know that is what the `subprojects` is doing here, but.
Think we could git clone directly, and leave meson for next.

shawnl:
I was wondering if we should send spec files over the wire https://github.com/distcc/distcc/issues/253

espenfk:
Someone working on this issue now? In a large code base, not being able to use split-dwarf, distcc and ccache has two issues:
- The debugger won't work, since the dwo file is not returned to the client
- ccache doesn't find the dwo file (for the same season) and no cache entry is made for the remotely compiled file.

Will this pull request solve the problem, or is it only a step on the way?

shawnl:
@espenfk split-dwarf support is actually completely not needed. It was mainly added because the ld.bfd and ld.gold linkers were running out of virtual memory on 32-bit systems, however the new lld linker does not map the debug sections of the .o files into VM, and does gain anything from -fsplit-dwarf.

espenfk:
@shawnl Thank you for swift answer. There is one more issue for us, and that is disk space. On Jenkins we saved 1,5Gb just on one build with split-dwarf because the debug information was not duplicated through all the libraries and binaries. And with a lot of builds for various platforms and configurations, we save enormous amounts of disk space by using split-dwarf. Just a case in favor of  supporting split-dwarf in distcc ... :-)

shawnl:
>  On Jenkins we saved 1,5Gb just on one build with split-dwarf because the debug information was not duplicated through all the libraries and binaries. 

Just build everything shared and you avoid this problem. While this might be a problem for deployment, it should not be a problem for testing.

PR #238, opened 2017-11-05
Android support

shawnl:
Amazingly this what the only patch require to build on temux---NATIVELY

https://github.com/termux


shawnl:
I am kinda just throwing -pie in there. The native toolchain I was using knew to use it, but it was an issue when cross-compiling. The thing is that this detection code doesn't work when cross-compiling.

PR #243, opened 2017-11-22
access control on having masquerade set up for binary

shawnl:
fixes the very old CVE-2004-2687 https://github.com/distcc/distcc/issues/155
at least unless gcc or clang can be exploited into executing arbitrary code--
but that is much more difficult than the current exploit.
(C++ templates are turing complete [1]....)

[1] http://web.archive.org/web/20131101122512/http://ubietylab.net/ubigraph/content/Papers/pdf/CppTuring.pdf

needs something like debian's update-distcc-symlinks script to maintain the whitelist. https://sources.debian.net/src/distcc/3.1-6.2/debian/update-distcc-symlinks.in/

What happens:

distcc check if a symlink to the binary exists in /usr/lib/distcc
if not, close the connection

This does not work with absolute pathnames to compilers.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -18,12 +18,13 @@ QUICK SUMMARY
 
 1. Build and install
 
-      ./autogen.sh       # If "configure" does not already exist.
+      ./autogen.sh              # If "configure" does not already exist.
       ./configure
       make
-      make check         # Optional! Should have python >= 3.1 installed.
-      make install       # You may need to use "sudo" for this command.
-      make installcheck  # Optional! Should have python >= 3.1 installed.
+      make check                # Optional! Should have python >= 3.1 installed.
+      make install              # You may need to use "sudo" for this command.
+      make installcheck         # Optional! Should have python >= 3.1 installed.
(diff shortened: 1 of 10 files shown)


shawnl:
going to need the script in order to pass the test suite

shawnl:
@afbjorklund Can I pack an error to the client through SOUT?

shawnl:
The question with the compiler detection script, is what language to write it in?

thedrow:
LGTM. Anyone else?

shawnl:
@thedrow Without a compiler detection script, this will break everything unless --make-me-a-botnet is passed, or if masquerade is set up. On Debian that script is already shipped and ran, so it breaks much less stuff.

shawnl:
Also, would be nice if a better error message could be presented to the user, by presenting on STDERR along with a failure, perhaps?

shawnl:
I added an error message, but if I return anything then distcc retry's locally and then complains that the results don't match. So a nice error message can only really happen my protocol version 4 patch set https://github.com/distcc/distcc/pull/236

PR #244, opened 2017-11-26
cross-compilation support

shawnl:
rewrites calls gcc to use fully qualified names
(like x86_64-linux-gnu-gcc), and fills in clang's -target option if it is not already filled in.


shawnl:
@stephengroat might you know why the second patch (rewriting "cc") fails on OSX? I don't have access to a mac.

PR #249, opened 2017-12-30
Create Brewfile

stephengroat:
moves osx dependencies out of `.travis.yml` and prevents errors from `brew install` if package is already installed (i.e. if travis changes the base `osx` image)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,7 @@ addons:
       - libiberty-dev
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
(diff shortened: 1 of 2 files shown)


stephengroat:
@shawnl easy to add to `osx` via `brew`, not as easy (but still possible) to add to `linux` since `meson` isn't available in the default `apt` repos, once it gets merged i can add it in to `osx` quickly

stephengroat:
any interest in this at all?

shawnl:
Why is the CI failing?

stephengroat:
Sorry about that, needed a new macOS version

Also added in a bunch of the newer ubuntu images (`xenial: clang` fails with a newer error check, just left it in under `allow_failures`)

PR #250, opened 2018-01-10
drop VOS support

mostynb:
I strongly suspect that this is untested/unused.

### DIFF
--- a/src/ncpus.c
+++ b/src/ncpus.c
@@ -64,32 +64,6 @@ int dcc_ncpus(int *ncpus)
 }
 
 
-#elif defined(__VOS__)
-
-#ifdef __GNUC__
-#define $shortmap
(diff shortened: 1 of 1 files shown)


shawnl:
I ported distcc to meson build system, which would mean dropping old platforms like this. The problem is that there are few programmers working on distcc. To review patches and the like.

thedrow:
I have no idea what VOS actually is.

shawnl:
It is an operating system.

thedrow:
According to wikipedia it is still active. Worth dropping support if there's any benefit to it.
@shawnl How does the code bother you?

shawnl:
Not my bug just answering the question. Doesn't bother me.

shawnl:
no need to remove this

PR #252, opened 2018-02-05
support some of gcc's most sensible -x options

shawnl:
Includes objective-c and go support


shawnl:
This patch is needed (at least) for x code support.

thedrow:
Thanks!

PR #261, opened 2018-02-28
Enabled ccache so that the builds would run faster.

thedrow:


thedrow:
The next build should finish faster (on Linux)...

shawnl:
IMHO the builds happen fast enough

thinrope:
Did it help and by how much?
Looking at https://travis-ci.org/distcc/distcc/builds there is no siginificant change...
I guess the cache (of ccache) is dropped with each build, so unless we have multiple builds (e.g. with different compile options) in a single travis build, it makes no sense.
Just my thoughts, might be completely wrong :-|

thedrow:
@thinrope Since there is no other commit in this PR there's no way to measure. The first time the build runs populates the cache. After that the cache will be reused.

PR #266, opened 2018-03-03
3.3

shawnl:


afbjorklund:
@shawnl : something like this: 56e72626c293d92bfe1345894d042265f049ede9

``` make
update-distcc-symlinks: update-distcc-symlinks.py
        cp $< $@

```

shawnl:
Cygwin needs the extention

shawnl:
But thanks for spotting rrhe problem.

shawnl:
I deleted the release until this gets fixed, perhaps without cygwin support.

afbjorklund:
Should be fine with Cygwin (which is a Unix), but OK - I can see how it can be an issue with Windows.
So it would need something similar to `@EXEEXT@`, in order to work with good old DOS extensions ?

I was wondering what include-server does, but apparently it uses `$(PYTHON)` to do the build/install.
Maybe something similar can be done even for a single file, otherwise I will just see what I can find...

afbjorklund:
Hmm, plot thickens... Without pump mode, we don't even _look_ for python(3).
`PYTHON=''`

If we now have a mandatory program using it, we should probably look always.

afbjorklund:
Doesn't build for Windows anyway, so might as well leave it (without suffix)

```
checking for x86_64-w64-mingw32-gcc... x86_64-w64-mingw32-gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.exe
checking for suffix of executables... .exe
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether x86_64-w64-mingw32-gcc accepts -g... yes
checking for x86_64-w64-mingw32-gcc option to accept ISO C89... none needed
checking for socklen_t... no
checking for socklen_t equivalent... configure: error: Cannot find a type to use in place of socklen_t
```

shawnl:
What Windows was this. and was it under cygwin's terminal? I had no problem with cygwin/windows7.

afbjorklund:
It was MinGW (https://mingw-w64.org). As in not Cygwin, just Windows...

But anyway, even though `python.exe` has an extension, that does not
mean that the scripts need to - since they are handled by shell/shebang:
`#!/usr/bin/env python3`

PR #267, opened 2018-03-05
Fix installation of update-distcc-symlinks

afbjorklund:
You will need this, since `update-distcc-symlinks.py` is missing from the distribution tarball


shawnl:
I just installed distcc on cygwin/windows, with this patch. The only problem is that /usr/sbin and /usr/local/sbin are not in the PATH variable by default.

shawnl:
windows 7

Issue #253, opened 2018-02-16
flag to allow -spec arguments


digitalist:
When trying to use distcc to build deb packages on identical machines with identical gcc setup, distcc works only on localhost if it sees gcc -spec files arguments (even with clang, though clang doesn't use this arg)
Since the env on all machines is identical, spec files on them are exact copies. It would be nice to have some option to add exclusions to distcc decision process, such as --allow-distributed-args="spec,some_other_flags"

shawnl:
The code is in src/arg.c

Distcc hasn't had a release in years, an important patch was not part of that release too....(but Debian carries it).

I will review a patch.

digitalist:
Hope you won't feel bad after reviewing this code, I'm the worst C programmer on this side of the ocean.

1. I moved logic responsible for checking possible flags from distcc_check_args to new function: `check_army_of_clones`, there is only one (`spec`) working flag for now, but if someone will want to add additional exlcusions,  it's better to have relevant code in one place

2. This function gets envvar `DISTCC_AOC` as a list of exclusions, and allows to distribute payload if it 
finds relevant exclusion in that var

ie: this will result in distributed build
 `export DISTCC_AOC="spec"  ./distcc c++ helloworld.cpp -specs=123` 

this will result in 'must be local' message in logs
`export DISTCC_AOC=""  ./distcc c++ helloworld.cpp -specs=123` 


https://github.com/digitalist/distcc/commit/70a627bdb5a22b09adcd28372285055daf3bd0e1

If your sanity will suffer after reviewing this code, I'm sorry.

PR #268, opened 2018-03-06
flag to allow -spec arguments

digitalist:
https://github.com/distcc/distcc/issues/253

I moved logic responsible for checking possible flags from distcc_check_args to new function: check_army_of_clones, there is only one (spec) working flag for now, but if someone will want to add additional exlcusions, it's better to have relevant code in one place

This function gets envvar DISTCC_AOC as a list of exclusions (delimited with ,), and allows to distribute payload if it
finds relevant exclusion in that var

ie: this will result in distributed build
export DISTCC_AOC="spec" ./distcc c++ helloworld.cpp -specs=123

this will result in 'must be local' message in logs
export DISTCC_AOC="" ./distcc c++ helloworld.cpp -specs=123


shawnl:
I feel the real solution is to change the protocol (put this in protocol very. 4 that is in development) and ship the spec file over the wire. Thoughts?

You can fix the build by rebasing with master.

afbjorklund:
What's with the "army of clones" thing ? Cutesy ?

digitalist:
@shawnl I guess it depends on a speed with which v4 is developed? If it's a year, I'd vote for including this patch, if it's half a year - we should throw it away:)

@afbjorklund something like this, yes.Sounds better then 'identical_set_of_building_hosts'

shawnl:
The problem with allowing specs files is that we don't have a sanitizer for them, for at least -fplugin=. see https://github.com/distcc/distcc/pull/276 and https://youtu.be/bSkpMdDe4g4?t=53m12s

digitalist:
well, this option is definitely meant for trusted env, trusted source and manual enabling. Maybe we should add an option to disable some risky variants, but idk if it will ever be enough to compile random code on untrusted machines

PR #271, opened 2018-03-12
Add Code Helpers Badge

TafThorne:
The CodeTriage website (https://www.codetriage.com/) offers a simple way to encourage more people to actively participate in an Open Source project.  People have an option to sign up and get emailed a suggestion of N issues to work on periodically.  Adding a badge about the service to the ReadMe.md will display the service clearly.


sourcefrog:
I'm sure they mean well at CodeTriage.org but I have a couple of concerns:

1. The track record of encouraging new developers to start on random issues is not good: many issues in all projects are not easily accessible to people who are new to the project, and some of these projects just create more noise for maintainers. In particular they are encouraging new people to triage issues, but triage often requires a fair amount of context.
2. For distcc in particular, which has been neglected for a while, I would welcome patches that fix real problems or to know when people can reproduce or diagnose bugs but I really don't want noise.

PR #278, opened 2018-03-31
Strip compiler arguments starting with '-isystem'

mgorny:
The chromium build system (using clang) apparently concatenates
'-isystem' with path (i.e. uses it without space). Support stripping
this version as well to avoid the lot of warnings:

    x86_64-pc-linux-gnu-clang++-6.0: warning: argument unused during compilation: '-isystem /usr/include/nss' [-Wunused-command-line-argument]


shawnl:
Wait, did you test this with pump mode?

mgorny:
Nope, sorry. Last I tried pump mode, it was completely broken and miscompiled stuff hard. Didn't ever step anywhere near it since then.

PR #285, opened 2018-06-07
Various patches mostly for cross-platform (Mac) support

RJVB:
This PR contains 5 improvements and fixes (in 6 distinctive commits):

- prevent a nullptr dereference if Avahi client creation fails
- support systems that don't have fstatat() and family; provide a fallback compiler whitelist check
  that doesn't require faccesasat()
- the `--allow-private` option doesn't take an argument
- don't hardcode `/usr/lib/distcc` but `$PREFIX/distcc` instead
- fix  the QtCreator support workaround

The fstatat/faccessat chance was made in 2 commits; initially I thought it sufficient to skip the whitelisting feature on the concerned legacy systems but then I realised it was trivial to add an equivalent fallback implementation not using `faccessat()`. I didn't manage to squash the 2 commits without altering other commits.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -383,7 +383,7 @@ AC_CHECK_FUNCS([strndup strsep mmap strlcpy])
 AC_CHECK_FUNCS([getloadavg])
 AC_CHECK_FUNCS([getline])
 
-AC_CHECK_FUNCS([fstatat],[],[AC_MSG_WARN([fstatat() not available; distccd will be slightly less secure])])
+AC_CHECK_FUNCS([fstatat])
 
 AC_CHECK_DECLS([snprintf, vsnprintf, vasprintf, asprintf, strndup])
 
--- a/configure.ac
+++ b/configure.ac
@@ -383,6 +383,8 @@ AC_CHECK_FUNCS([strndup strsep mmap strlcpy])
 AC_CHECK_FUNCS([getloadavg])
 AC_CHECK_FUNCS([getline])
 
+AC_CHECK_FUNCS([fstatat],[],[AC_MSG_WARN([fstatat() not available; distccd will be slightly less secure])])
+
 AC_CHECK_DECLS([snprintf, vsnprintf, vasprintf, asprintf, strndup])
 
 AC_MSG_CHECKING([if mmap() supports MAP_FAILED])
--- a/src/dopt.c
+++ b/src/dopt.c
@@ -135,7 +135,7 @@ static const char *dcc_private_networks[] = {"192.168.0.0/16",
 
 const struct poptOption options[] = {
     { "allow", 'a',      POPT_ARG_STRING, 0, 'a', 0, 0 },
-    { "allow-private", 0,POPT_ARG_STRING, &opt_allow_private, 0, 0, 0 },
+    { "allow-private", 0,POPT_ARG_NONE, &opt_allow_private, 0, 0, 0 },
 #ifdef HAVE_GSSAPI
     { "auth", 0,	 POPT_ARG_NONE, &opt_auth_enabled, 'A', 0, 0 },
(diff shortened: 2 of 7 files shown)


configure.ac
```diff
@@ -383,6 +383,8 @@ AC_CHECK_FUNCS([strndup strsep mmap strlcpy])
 AC_CHECK_FUNCS([getloadavg])
 AC_CHECK_FUNCS([getline])
 
+AC_CHECK_FUNCS([fstatat],[],[AC_MSG_WARN([fstatat() not available; distccd will be slightly less secure])])
```
shawnl: what systems don't have fstatat?

https://gist.github.com/Birch-san/ebca41072b8ade2be0bfbc18b110fb94

src/dopt.c
```diff
@@ -135,7 +135,7 @@ static const char *dcc_private_networks[] = {"192.168.0.0/16",
 
 const struct poptOption options[] = {
     { "allow", 'a',      POPT_ARG_STRING, 0, 'a', 0, 0 },
-    { "allow-private", 0,POPT_ARG_STRING, &opt_allow_private, 0, 0, 0 },
+    { "allow-private", 0,POPT_ARG_NONE, &opt_allow_private, 0, 0, 0 },
```
shawnl: thank you

shawnl:
I havn't had good success with the botnet option, but the old code was really insecure, and the CMDLIST stuff was too difficult to set-up. What do you recommend on the future direction of this code?

src/serve.c
```diff
@@ -375,13 +375,20 @@ static int dcc_check_compiler_whitelist(char *_compiler_name)
     /* Support QtCreator by treating /usr/bin and /bin absolute paths as non-absolute
      * see https://github.com/distcc/distcc/issues/279
      */
-    if (strstr(_compiler_name, "/bin/"))
-        compiler_name = _compiler_name + strlen("/bin/");
-    else if (strstr(_compiler_name, "/usr/bin/"))
-        compiler_name = _compiler_name + strlen("/usr/bin/");
+    const char *creator_paths[] = { "/bin/", "/usr/bin/", NULL };
+    for (int i = 0 ; creator_paths[i] ; ++i) {
```
shawnl: Fix the C89 error here

RJVB:
>+AC_CHECK_FUNCS([fstatat],[],[AC_MSG_WARN([fstatat() not available; distccd will be slightly less secure])])
>
>what systems don't have fstatat?

Older Mac OS (X) (before 10.11 IIRC). Given those are based on BSD there are probably *BSD version for which this is true also.

>I havn't had good success with the botnet option, but the old code was really insecure, and the CMDLIST stuff was too difficult to set-up. What do you recommend on the future direction of this code?

I cannot really answer that; I haven't looked at this mode at all, not even what it's supposed to do. 
Isn't it in fact a mode in which there's basically no security at all in the sense that distccd acts as a distributed any-kind-of-job server?

I can see how this could be useful for certain users, and I think there's nothing inherently wrong with an insecure mode if you have to activate it deliberately and supposedly with full knowledge of the potential consequences and/or how to protect the (distributed) system.
You could make it a build option to have the mode available so it can really not be activated by accident.

FWIW, my initial POV concerning whitelisting (in absence of faccessat) was that the many if not most users will probably be distributing jobs on a protected/private LAN anyway, meaning that threats could only come from a limited number of known machines. If such a LAN ever gets compromised you probably have more urgent things to worry about than distcc exploits.

RJVB:
Great, happy to see my efforts were of help.

Any idea when the next release will land, or alternatively, how well suited for "production use" the current version is? 
I haven't noticed any issues myself (deploying distccd on a Mac with cycles to spare, to the benefit of a much slower Linux machine) but I haven't looked too hard either.

PR #313, opened 2018-11-09
Handle complex arguments, like assembler macros

### LINKED ISSUE #312: Linux Kernel Assembly Macros
Hi,

As of v4.20-rc1 distcc (and icecc for that matter) can no longer be used to distribute a compile of the linux kernel for x86. See this [thread](https://lore.kernel.org/lkml/89262174-6f15-6116-e67f-436f7c1bfe99@deltatee.com/T/#u).

The problem is the kernel is now adding "-Wa,macros.s" argument to every compile option so that specific assembly macros can be used in inline assembly code across the kernel. When distcc sees this, any remote builds will fail seeing it won't be able to find macros.s and it will fallback to a local build.

Is there any way this could be supported by distcc as there does not appear to be a good solution in the kernel?

Thanks,

Logan

afbjorklund:
The Linux kernel has started using statements like
-Wa,macros.s for including assembler with the code.

Handle these similar to the `-imacros` statement,
and send them to the compile server for processing.

Closes #312

### DIFF
--- a/include_server/parse_command.py
+++ b/include_server/parse_command.py
@@ -119,6 +119,22 @@ CPP_OPTIONS_MAYBE_TWO_WORDS_FIRST_LETTERS = ('M', 'i', '-', 'a', 't')
 for key in CPP_OPTIONS_MAYBE_TWO_WORDS.keys():
   assert key[1] in CPP_OPTIONS_MAYBE_TWO_WORDS_FIRST_LETTERS
 
+PATH_EXPR='[/a-zA-Z_0-9.]+' # regular expression for a partial file path
+
+# These are the cpp options that require regular expressions, m is Match.
+CPP_OPTIONS_REGULAR_EXPRESSIONS = {
+  '-Wa,(%s\.s)' % PATH_EXPR:     lambda ps, m: ps.include_files.append(m.group(1)),
+  '-Wa,\[(%s\.s)\]' % PATH_EXPR: lambda ps, m: ps.include_files.append(m.group(1)),
+}
+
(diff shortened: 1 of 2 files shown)


include_server/parse_command.py
```diff
@@ -119,6 +119,24 @@ def _RaiseNotImplemented(name, comment=''):
...
 
+PATH_EXPR='[/a-zA-Z_0-9.]+' # regular expression for a partial file path
+
+# These are the cpp options that require regular expressions, m is Match.
+CPP_OPTIONS_REGULAR_EXPRESSIONS = {
+  '-Wa,(%s\.s)' % PATH_EXPR:     lambda ps, m: ps.include_files.append(m.group(1)),
+  '-Wa,\[(%s\.s)\]' % PATH_EXPR: lambda ps, m: ps.include_files.append(m.group(1)),
+  '-Wa,(%s\.S)' % PATH_EXPR:     lambda ps, m: ps.include_files.append(m.group(1)),
+  '-Wa,\[(%s\.S)\]' % PATH_EXPR: lambda ps, m: ps.include_files.append(m.group(1)),
+}
```
lsgunth: It may not be necessary to support .S files as, by convention, they need to be run through the preprocessor and the -Wa argument will skip the pre-processor. 

In the kernel, macro.s is already generated from macro.S in a different step.

lsgunth:
I tested this on my setup and it seems to work.

include_server/parse_command.py
afbjorklund: Good to know, we do run cpp on the server (in pump mode) but if it doesn't handle it anyway...

PR #315, opened 2018-11-20
Improve Dockerfile

PeterDaveHello:


TafThorne:
These changes look like general good practice for dockerfile work so they get a +1 from me.  I have not yet had time to verify that the dockerfiles work correctly, or I would approve this change.  I do not wanto assume that we are automatically building the docker file in the passed checks.  

Thank you for your contribution PeterDaveHello.

PeterDaveHello:
Hi @shawnl, do you think we have a chance to get this merged? Thanks!

PeterDaveHello:
It's sad that almost 4 years passed and we don't have any updates here 🤣

PeterDaveHello:
@fergushenderson @sourcefrog hope you guys can help take a look 😅 Thanks!

docker/base/Dockerfile
```diff
@@ -7,9 +7,11 @@ ENV DEBIAN_FRONTEND=noninteractive
...
 RUN apt-get update && \
-    apt-get install autoconf \
-                    python3-dev \
-                    python-dev \
-                    libiberty-dev \
-                    build-essential \
-                    make
+    apt-get install -y autoconf \
+                       python3-dev \
+                       python-dev \
```
andy5995, line 12: I doubt this line is needed. The correct deps should be installed when python3-dev is installed.

andy5995 (commented):
xenial is no longer supported by Ubuntu, should this Dockerfile be importing bionic instead?

docker/base/Dockerfile
PeterDaveHello: @andy5995 thanks for the input, this PR is originally for the Dockerfile basic improvement, using the most common best practices, so other deep-dive improvement were not considered to be together at that moment. Just FYI ;)

sourcefrog:
Hi, I'm not active in this any more, but the changes look reasonable. Can you update it to a current OS base image? Otherwise good to go.

docker/base/Dockerfile
andy5995: @PeterDaveHello Understood :)

PR #319, opened 2018-12-08
Fix desktop spec compliance of distccmon-gnome install

mgorny:
Update the .desktop file and install rules for distccmon-gnome to
conform to the freedesktop.org specifications.  More specifically:

- update to version 1.0 of the spec
- remove obsolete 'Encoding' key
- remove obsolete 'Application' category and add recommended 'GTK'
- install icon file to the correct directory
- strip suffix from desktop file (specifying it is deprecated)
- while at it, remove redundant '-icon' from the filename
- install desktop file to the correct directory


rossburton:
I was about to do this, so :+1: from me.

PR #324, opened 2019-02-06
When waiting for the wrapped process to exit, make sure it's not a zombie

Thermi:
Makes pump work inside docker when used by my docker-makepkg script. Otherwise it deadlocks.
The patch does exactly what is described in the comment.


afbjorklund:
Installing an `init` process in your container (e.g. `docker run --init`) should help against zombies.

Thermi:
Merging the commit solves the problem generally. Then it's not required to add exceedingly more complexity to the container at runtime.

PR #335, opened 2019-04-14
Fix syntax error for uninstall-man

mako4th:


shawnl:
You might be interested in my meson patches  https://github.com/distcc/distcc/pull/236

I don't really expect the uninstall to work.

shawnl:
There will be no more development on the Makefile build system. It is deprecated for meson.

mako4th:
Ok, I understand.

PR #336, opened 2019-04-14
Fix memory leaks detected by PVS-Studio

Oleh-Kravchenko:
- climasq.c:121 The function was exited without releasing the 'buf' pointer.
- argutil.c:109 The function was exited without releasing the 'b' pointer.
- stringmap.c:53 The function was exited without releasing the 'result' pointer.
- util.c:596 when realloc() fails in allocating memory,
  original pointer 'loc' is lost.

Signed-off-by: Oleh Kravchenko <oleg@kaa.org.ua>

### DIFF
--- a/src/stringmap.c
+++ b/src/stringmap.c
@@ -43,14 +43,20 @@
  */
 stringmap_t *stringmap_load(const char *filename, int numFinalWordsToMatch)
 {
-    stringmap_t *result = calloc(1, sizeof(*result));
-    FILE *fp = fopen(filename, "r");
+    stringmap_t *result;
+    FILE *fp;
     char buf[2*PATH_MAX];
     int n;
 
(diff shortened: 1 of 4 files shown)


src/stringmap.c
```diff
@@ -43,14 +43,20 @@
...
+    FILE *fp;
     char buf[2*PATH_MAX];
     int n;
 
+    result = calloc(1, sizeof(*result));
+    if (!result)
+        return NULL;
     result->numFinalWordsToMatch = numFinalWordsToMatch;
-    if (!fp)
+    fp = fopen(filename, "r");
```
shawnl: you can try to open the file first, so that you don't have to free.

shawnl: https://github.com/distcc/distcc/commit/bbfddaa66e85913abce5f978c6b22a11d37326bc

PR #345, opened 2019-06-11
Zstd with meson

shawnl:
see https://github.com/distcc/distcc/pull/236

The distcc package in Debian is in a really poor state, and switching the meson will force the maintainer there to fix the problems.

This should also fix the reports of people who do not enable rfc2553 and then complain about ipv6 problems.

Yes, zstd is not optional, but it still supports the old protocols.


afbjorklund:
There seems to be a _lot_ going on at once, in this PR ?

build systems, python versions, compression algorithms

shawnl:
> python versions, 

I wasn't trying to change this.

If I can't get the CI running smoothly there is no chance I will merge this anyways. The zstd code seemed to work fairly well.

shawnl:
As mentioned above, the main reason I have for releasing a new version is to force the Debian packaging to improve. My bug reports have proven ineffectual, and the maintainer is not shipping the upstream version of `update-distcc-symlinks` because of a bug he didn't have the energy to investigate.

PR #349, opened 2019-06-14
Fixes

shawnl:


src/where.c
```diff
@@ -173,7 +173,7 @@ static int dcc_lock_one(struct dcc_hostdef *hostlist,
     int ret;
 
     while (1) {
-        for (i_cpu = 0; i_cpu < 50; i_cpu++) {
+        for (i_cpu = 0; i_cpu < 50000; i_cpu++) {
```
afbjorklund: Seems to be changing one arbitrary number for another ? Think I liked the suggestion in the bug report better

shawnl: My understanding of this code was this it also locks later, so this lock is not really necessary.

Issue #331, opened 2019-03-20
Better handling of -march=native


davidrohr:
Currently, when the command line contains -march=native (or -mtune=native), distcc will not distribute to other nodes but compile locally.

This could by improved to replace the -march=native by the corresponding arguments to which -march=native translates with the local compiler.

I had a look and I think this can be implemented in arg.c in dcc_scan_args function.
It has access to argv[0], so it knows which compiler to use, and it already creates new arguments in in ret_newargv. It could easily replace -march=native with the proper settings for the local architecture. These settings could either be determined automatically using argv[0] and invoking the compiler once, ideally caching the result. Or they could be specified in the config file.

I would be willing to contribute such a patch, but before I was wondering if this would be accepted in the end, or whether there are already plans to implement something like this, and whether the approach I suggest sounds reasonable.

shawnl:
If you get the flags from invoking the compiler, and cache the result using the back-off code (when a host is broken) I will review it, however I don't see this feature as very useful, and I like how stateless distcc is, which this reduces.

XenHat:
This is something I would also be interested in getting. Making distcc parse the result of `gcc -### -E - -march=native 2>&1 | sed -r '/cc1/!d;s/(")|(^.* - )//g'`
(suggested from https://stackoverflow.com/a/18963118) and send that to the contributing machines should work.

On my machine, this equates to:
```
$ gcc -### -E - -march=native 2>&1 | sed -r '/cc1/!d;s/(")|(^.* - )//g'
[...]

PR #350, opened 2019-06-18
Add feature to automatically resolve -march/mtune=native flags

davidrohr:
As discussed in #331 , this PR adds a feature which automatically resolves the necessary architecture flags for march-native with gcc and clang, and passes the resolved flags to the compiler.

The feature is disabled by default, and is activated with an environment variable.

Initially I wanted to cache the result, but then I decided this is probably not ideal, since the compiler may be a symlink that changes (or alike), so caching seems dangerous.
Resolving every time seems not to introduce a large overhead for me.

### DIFF
--- a/src/arg.c
+++ b/src/arg.c
@@ -110,6 +110,164 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
            "compile from %s to %s", input_base, output_base);
 }
 
+/* Look for march/mtune=native in the arguments, and replace with the
+ * actual arguments employed by gcc / clang. */
+static int dcc_resolve_march_native(char* argv[], char*** ret_newargv, int extra_args, int* ignore_range_min, int* ignore_range_max)
+{
+    int i, l, j;
+    l = dcc_argv_len(argv);
+    if (l == 0) return 0;
+    if (!argv[0]) return 0;
+    const char* compiler = strrchr(argv[0], '/');
+    compiler = (compiler == 0) ? (argv[0]) : (compiler + 1);
+    int is_clang = strncmp(compiler, "clang", strlen("clang")) == 0;
+    int found_arch_native = 0, found_tune_native = 0, found_cpu_native = 0;
+    for (i = 0;i < l;i++) {
+        if (!strcmp(argv[i], "-march=native")) {
+            found_arch_native = 1;
+        }
+        if (!strcmp(argv[i], "-mtune=native")) {
+            found_tune_native = 1;
+        }
+        if (!strcmp(argv[i], "-mcpu=native")) {
+            found_cpu_native = 1;
+        }
+    }
+    if (!found_arch_native && !found_tune_native && !found_cpu_native) return 0;
+
+    rs_trace("-march/-mtune/-mcpu=native option found, replacing...");
+
+    char buff[16384];
+
+    int outfds[2];
+    int infds[2];
+    int ret = pipe(outfds);
+    if (ret == -1) {
+        fprintf(stderr, "Failed to open stdout pipe: %s (%d)\n", strerror(errno), errno);
+        return 1;
+    }
+    ret = pipe(infds);
+    if (ret == -1) {
+        fprintf(stderr, "Failed to open stdin pipe: %s (%d)\n", strerror(errno), errno);
+        return 1;
+    }
+    int outRead = outfds[0];
+    int outWrite = outfds[1];
+    int inRead = infds[0];
+    int inWrite = infds[1];
+
+    switch (fork()) {
+    case -1:
+        fprintf(stderr, "Failed to fork: %s (%d)", strerror(errno), errno);
+        return 1;
+    case 0: { // Child
+        close(outWrite);
+        dup2(outRead, STDIN_FILENO);
+        close(outRead);
+        close(inRead);
+        dup2(inWrite, STDERR_FILENO);
+        close(inWrite);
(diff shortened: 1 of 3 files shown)


src/arg.c
```diff
@@ -110,6 +110,111 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
...
+        }
+    }
+    if (!found_native) return 0;
+    char buff[16384];
+    FILE *in;
+    if (!argv[0]) return 0;
+    snprintf(buff, sizeof(buff),
+        "%s -v -E -x c -march=native -mtune=native - < /dev/null 2>&1", argv[0]);
+    if(!(in = popen(buff, "r")))
+    {
```
shawnl: Can you please match the style in the rest of the code.

src/arg.c
```diff
@@ -110,6 +110,111 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
...
 }
 
+/* Look for march/mtune=native in the arguments, and replace with the
+ * actual arguments employed by gcc / clang. */
+static int dcc_resolve_march_native(char* argv[], char*** ret_newargv, int extra_args, int* ignore_range_min, int* ignore_range_max)
+{
+    int i, l, j;
+    l = dcc_argv_len(argv);
+    if (l == 0) return 0;
+    const char* compiler = strrchr(argv[0], '/');
```
shawnl: This is going to require switching to c11 mode (which is default in all compilers except the ancient one that travisCI uses), but that is OK.

src/strip.c
```diff
@@ -111,7 +119,7 @@ int dcc_strip_local_args(char **from, char ***out_argv)
                  || str_startswith("-isystem", from[from_i])) {
             /* Something like "-DNDEBUG" or
              * "-Wp,-MD,.deps/nsinstall.pp".  Just skip this word */
-            ;
+            continue;
```
shawnl: These two are unrelated, but they seem correct.

src/arg.c
```diff
@@ -110,6 +110,111 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
...
+            break;
+        }
+    }
+    if (!found_native) return 0;
+    char buff[16384];
+    FILE *in;
+    if (!argv[0]) return 0;
+    snprintf(buff, sizeof(buff),
+        "%s -v -E -x c -march=native -mtune=native - < /dev/null 2>&1", argv[0]);
+    if(!(in = popen(buff, "r")))
```
shawnl: Can you please do this without using a shell. I know that is kind of a PITA in C, but adding a dependency on the shell is unacceptable.

davidrohr: ok, I can have a look, but actually I have no idea how to do it without a shell

shawnl: https://git.dpkg.org/cgit/dpkg/dpkg.git/tree/dpkg-deb/build.c#n435

sandsmark: sorry for the horrible code (and I mangled a bit of your code just to test it easily), but here's one approach using pipe(), dup2 (to replace stdin and stderr for the compiler) and execlp (just because it was easier for testing, not sure if that's easier here, but should be trivial to replace): https://gist.github.com/sandsmark/9f4cc47b6cf4b11f05d677624e0ae025

davidrohr:
I finally found some time to work on this:
- some fixes + separate treatment of march / mtune / mcpu
- Included the code from @sandsmark to get the arguments without using a shell (thx!, saved me quite some time)
- Fixed some style inconsistencies (I am not 100% sure what the style should be, even the file before my changes is not 100% consistent. If you require more changes, please let me know, or just push to the branch, it is editable by maintainers.)

Thermi:
Kindly requesting an update @shawnl @afbjorklund

davidrohr:
Currently it seems to break the CI, needs some minor adjustment for the old compilers.

themightyoarfish:
Any chance on revisiting this?

davidrohr:
in principle yes, it works, just needs to be slightly adapted to pass the continuous integration...

davidrohr:
I have fixed the C99 errors. Let's see if the CI passes now.

davidrohr:
@shawnl : could you trigger the CI again please?

src/arg.c
```diff
@@ -135,9 +293,17 @@ int dcc_scan_args(char *argv[], char **input_file, char **output_file,
     char *a;
     int ret;
 
+    int ignore_range_min = 0, ignore_range_max = 0;
+    if (getenv("DISTCC_RESOLVE_MARCH_NATIVE") &&
```
zougloub, line 297: Add documentation about the new environment variable (`man/distcc.1` § `.SH "ENVIRONMENT VARIABLES"`)?

In the last paragraph of DESCRIPTION, I'd add a sentence about tasks that are run locally (preprocessing, linking, and `-m{arch,tune,cpu}=native` unless `$DISTCC_RESOLVE_MARCH_NATIVE` is set).

In CROSS COMPILING another sentence could be added to document the fact that -march=native is handled by the new environment variable.

You could also give different meanings to the `getenv()` value in order to do either:
- "forward": forward native argument to remote machine (dangerous unless remote has same processor)
- "explicit": resolve actual flags by invocation of local compiler, then attempt remote compilation with resolved flags; this incurs a subprocess call penalty.
- "local": just compile locally
- "fail": refuse to process a command line containing these arguments

I am not sure which default would be the sanest.
The best, for a user who wants speed-up, is for them to fixate the equivalent options.

Issue #352, opened 2019-06-27
distccd parent process dies when handling jobs while child processes are starting


jansmets:
We have some 4-way xeon machines with a large amount of cpu cores.

The first thing we observe is that distccd takes a long time to start. There is a hard coded sleep(1) for every child server, so it takes several minutes to start the service.

While the distccd service is starting, clients can send jobs and those get handled by the daemon. However, if we send a large amount of jobs while starting - more jobs than there are currently child processes running- then we see the parent process terminates with a segmentation fault. The childs (I think) are exiting with '(dcc_writex) ERROR: failed to write: Broken pipe'

```
(gdb) info line *0x4067fc
Line 179 of "src/stats.c" starts at address 0x4067fc <dcc_stats_server+2220> and ends at 0x406803 <dcc_stats_server+2227>.

168     /* drop elements older than 2min */
169     curr_sd = dcc_stats.sd_root;
170     while (curr_sd != NULL) {
171         if (curr_sd->stop.tv_sec < two_min_ago) {
172             /* delete the stat */
173             if (prev_sd == NULL) {
174                 dcc_stats.sd_root = curr_sd->next;
175             } else {
176                 prev_sd->next = curr_sd->next;
177             }
178             free(curr_sd);
179             curr_sd = prev_sd->next;
180         } else {
181             /* we didn't delete anyting. move forward by one */
182             prev_sd = curr_sd;
183             curr_sd = curr_sd->next;
184         }
185     }
186 }
```

To reproduce you can have a client with dummy compiles, eg
```
$ cat /tmp/test.c
int main(void) { return 0; }
$ mkdir /tmp/distcc-test
$ for i in $(seq 1 100); do cp /tmp/test.c /tmp/distcc-test/$i.c ; done
$ cd /tmp/distcc-test
$ cat Makefile
objects := $(patsubst %.c,%.o,$(wildcard *.c))
all: $(objects)
DISTCC:=distcc
%.o : %.c
        DISTCC_HOSTS="myserver:40000/200 --randomize " DISTCC_IO_TIMEOUT=1200 DISTCC_SKIP_LOCAL_RETRY=1 DISTCC_BACKOFF_PERIOD=0 $(DISTCC) gcc -c $< -o $@

on "myserver" run the daemon:
 /tmp/distccd  --stats --log-level=critical  --allow=0.0.0.0/0 --job-lifetime=1200 --stats-port 40001 --port 40000 --verbose --no-detach --log-stderr --jobs=125

[...]

PR #353, opened 2019-06-27
Fork childs faster

jansmets:
Work around for https://github.com/distcc/distcc/issues/352


sourcefrog:
@mattst88 points out in https://github.com/distcc/distcc/issues/352#issuecomment-593713941 that `sleep` in C takes an int not a float. Presumably this is getting rounded down at compile time to 0, presumably with a compiler warning. 

Probably the simplest thing here is to just remove the `sleep` calls. It's probably not a very effective protection against overloading the machine, as merely forking is cheap.

PR #355, opened 2019-07-14
Support GSSAPI auth with custom service principal name

rmatev:
Add an optional element to the hosts spec, `AUTH_NAME`, which is the "canonical" name to use for the service principal name instead of `HOSTNAME` (or its corresponding fqdn).

This option is necessary in case of accessing an authenticated server via ssh port forwarding, in which case the `HOSTNAME` is `127.0.0.`1 and the canonical name cannot be deduced automatically.

Also, fix a couple of issues revealed with GCC 8.3.1 and add some artefacts to `.gitignore`.

Testing of any authentication features is not done as far as I see, and requires a bit of setting up. I can take care of this if needed. In any case, I have been using this new feature for a few weeks already without issues.


shawnl:
Actually, can we get a test for this?

rmatev:
> Actually, can we get a test for this?

I can propose some changes that exercise the `DISTCC_HOSTS` parsing (only when compiled using `--with-auth`). It won't be extremely pretty but should be okay.

Proper testing of the GSSAPI authentication would need a local test KDC server, and setting that and everything else will take a considerable amount of time.

What do you think?

shawnl:
El dom., 14 jul. 2019 9:08, Rosen Matev <notifications@github.com> escribió:

> Actually, can we get a test for this?
>
> I can propose some changes that exercise the DISTCC_HOSTS parsing (only
> when compiled using --with-auth). It won't be extremely pretty but should
> be okay.
>
> Proper testing of the GSSAPI authentication would need a local test KDC
> server, and setting that and everything else will take a considerable
> amount of time.
>
> What do you think?b
>
Yeh that is fair enough. I am surprised the gssapi still works given that
there are no tests however.

PR #362, opened 2019-08-19
Autoconf cleanup

ngie-eign:
This change does 2 things:
* Allows `autoreconf -i -s` to work out of the box without having to specify `-I m4`.
* Replaces `acx_pthread.m4` with a more supported copy from the autoconf-archive project named `ax_pthread.m4`.


ngie-eign:
@shawnl: ping? I'm still running into this issue on my Fedora 30 workstation:
```
$ autoreconf -s -i
$ ./configure
...
./configure: line 6498: ACX_PTHREAD: command not found
```

Issue #377, opened 2020-04-13
Distcc fails to cross-compile on PC platforms


Shorrer:
Distcc removes the "-pc-" part from GNU_HOST definition of the host when trying to rewrite compiler command to the full one. At least on my system compilers do have that "-pc-" suffix in their names, so trying to check execute permissions for "i686-linux-gnu-*" gets ENOENT which causes Distcc to not rewrite names and that breaks cross-compilation.
1. I am really not an expert on GCC, cross-compilation and all that, but what is the reason behind removal of "-pc-" suffix?  If guess-script said that this is i686-pc-linux-gnu then it most likely says the same thing for others.
2. In order to make Distcc work I propose removing this check for "-pc-" suffix (pull request will come a bit later)

shawnl:
Yes, distcc removes this because clang can't handle it, IIRC. I can't merge this if it breaks clang cross-compilation.

Shorrer:
https://clang.llvm.org/docs/CrossCompilation.html
^This article acknowledges presence of "-pc-" suffix in target triplet.
Still, I haven't tested it, and I've never used clang before, so I'll come back when I'll have this checked to be sure.
Thanks for quick reply!

shawnl:
Ahhh, ok, then it is for the names of the GCC cross compilers, which do not
have that, on Debian.

Shorrer:
Oh, so then it's Debian package maintainers who can patch the compiler search code into "compile.c" and for the rest of us there will be names with "-pc-" vendor suffix. I think that would be right – distribution-specific quirks should be maintained by distribution itself.
For your information, I've encountered this on Gentoo.

shawnl:
But it breaks it for other people. Gentoo is a system where everyone configures it differently. The only "solution" I see here is to scan the whole directory so that the vector suffix can be ignored, and that is hardly a solution at all. I think that what distcc is doing removing the vendor part is correct, and that you should add a symlink to your compiler without the vendor part (as distcc doesn't care about vendor, as the code is already proprocessed).

Sorry, but you need to provide gcc without a vendor name. Closing wont-fix.

mgorny:
I'm not going to ask you to be reasonable but I just want to note that I find it really wrong to do such hacks just because some distro known for overpatching everything made a meaningless rename, and then asking everyone else to make such stupid changes too.

vsyrjala2:
[...]

PR #378, opened 2020-04-13
Removed check for -pc- suffix when rewriting the compiler command

Shorrer:
Fixes issue #377


shawnl:
See #377 for reason that is being closed.

PR #384, opened 2020-06-18
Add feature to automatically resolve -march/mtune=native flags (rebased and fixed)

sandsmark:
https://github.com/distcc/distcc/pull/350, just rebased and (hopefully) fixed compatibility with the older compilers in the CI system.

### DIFF
--- a/src/arg.c
+++ b/src/arg.c
@@ -110,6 +110,162 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
            "compile from %s to %s", input_base, output_base);
 }
 
+/* Look for march/mtune=native in the arguments, and replace with the
+ * actual arguments employed by gcc / clang. */
+static int dcc_resolve_march_native(char* argv[], char*** ret_newargv, int extra_args, int* ignore_range_min, int* ignore_range_max)
+{
+    int i, l, j;
+    l = dcc_argv_len(argv);
+    if (l == 0) return 0;
+    if (!argv[0]) return 0;
+    const char* compiler = strrchr(argv[0], '/');
+    compiler = (compiler == 0) ? (argv[0]) : (compiler + 1);
+    int is_clang = strncmp(compiler, "clang", strlen("clang")) == 0;
+    int found_arch_native = 0, found_tune_native = 0, found_cpu_native = 0;
+    for (i = 0;i < l;i++) {
+        if (!strcmp(argv[i], "-march=native")) {
+            found_arch_native = 1;
+        }
+        if (!strcmp(argv[i], "-mtune=native")) {
+            found_tune_native = 1;
+        }
+        if (!strcmp(argv[i], "-mcpu=native")) {
+            found_cpu_native = 1;
+        }
+    }
+    if (!found_arch_native && !found_tune_native && !found_cpu_native) return 0;
+
+    rs_trace("-march/-mtune/-mcpu=native option found, replacing...");
+
+    char buff[16384];
+
+    int outfds[2];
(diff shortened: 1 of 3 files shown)


shawnl:
I am skeptical of calling the compiler like this every time. It is quite expensive. How about adding the flags to the error message that the user gets when they pass `-march=native`, so they can rerun the compilation with the right flags, but without all this overhead?

sandsmark:
> I am skeptical of calling the compiler like this every time. It is quite expensive. How about adding the flags to the error message that the user gets when they pass `-march=native`, so they can rerun the compilation with the right flags, but without all this overhead?

the point (for me) is to avoid having to fix all build systems I happen to come across, so I'd rather waste 10ms (probably less, the variation when testing with `time` is pretty high). :-)

shawnl:
Then how about pulling the `-march=native` flags from the environment, and using a wrapper to set that environment variable?

sandsmark:
The best would be if distcc didn't require hacks like that.

But I really don't understand the objection at all. The call doesn't happen if you don't set march/mtune/whatever to native.

And I can't find a testcase where the call is expensive, how are you testing?

src/arg.c
```diff
@@ -110,6 +110,162 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
...
+/* Look for march/mtune=native in the arguments, and replace with the
+ * actual arguments employed by gcc / clang. */
+static int dcc_resolve_march_native(char* argv[], char*** ret_newargv, int extra_args, int* ignore_range_min, int* ignore_range_max)
+{
+    int i, l, j;
+    l = dcc_argv_len(argv);
+    if (l == 0) return 0;
+    if (!argv[0]) return 0;
+    const char* compiler = strrchr(argv[0], '/');
+    compiler = (compiler == 0) ? (argv[0]) : (compiler + 1);
```
shawnl, line 122: why can't it just call `argv[0]` as-is?

src/arg.c
```diff
@@ -110,6 +110,162 @@ static void dcc_note_compiled(const char *input_file, const char *output_file)
...
+                            *ignore_range_min = j;
+                            const char* ptr_insert = args;
+                            int clang_force_next = 0;
+                            for (char* ptr = args;*ptr;ptr++) {
+                                if (*ptr == ' ' || *ptr == 0) {
+                                    if (*ptr == ' ') *(ptr++) = 0;
+                                    const char* insert = ptr_insert;
+                                    ptr_insert = ptr;
+                                    if (is_clang) {
+                                        if (strncmp(insert, "-target", strlen("-target")) == 0) {
```
shawnl, line 229: How can this possibly work when `-target` is set? Unless you know that the `-target` is the native target, there is nothing that can be done with `=native`.

src/arg.c
sandsmark: because of the comparison on the next line, I guess.

src/arg.c
sandsmark: This is to remove `-target` if `-march=native`, `-mtune=native` or `-mcpu=native` I guess. I assumed it was because -target is implicitly set by -mcpu/-march/-mtune, but please correct me if it's wrong (the man-page for clang is ... not very verbose on what `-target` does).

shawnl: If you remove `-target` then you break cross-compiling.

davidrohr: I would not bother for `-target` too much for code that is meant for march=native. In that case we cannot cross-compile anyway

PR #385, opened 2020-06-30
Only add -MMD for pump mode if not present already

rmatev:
Currently, distccd unconditionally adds `-MMD` to the compiler
invocation in pump mode. This leads to undesired behaviour if one
passes `-MD` to the compiler. For GCC, this leads to prefering `-MMD`,
which is more restrictive and might cause incorrect builds due to missed
system header dependensies. OTOH, Clang prefers `-MD`, which is safer
but emits the following warning:

```
clang-8: warning: argument unused during compilation: '-MMD' [-Wunused-command-line-argument]
```

The changes here fix this by only adding `-MMD` if neither `-MMD` nor
`-MD` is already present. Some exceptions in the tests, related to this bug,
are removed and the corresponding tests are now passing for Clang and
GCC.

### DIFF
--- a/src/serve.c
+++ b/src/serve.c
@@ -591,7 +591,9 @@ static int tweak_arguments_for_server(char **argv,
     if ((ret = dcc_convert_mt_to_dotd_target(*tweaked_argv, dotd_target)))
       return 1;
 
-    dcc_argv_append(*tweaked_argv, strdup("-MMD"));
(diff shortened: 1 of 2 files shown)


rmatev:
thank you, that was really fast!

shawnl:
You provided a comprehensive justification for the change.

PR #387, opened 2020-07-08
Fix compiler warnings on Cygwin/MSYS2

orgads:


shawnl:
What is the warning you are seeing? Why can't the code be fixed instead?

orgads:
Do you prefer it this way?

shawnl:
This is wrong. Look at the man page for these functions:

```
       These functions check whether c, which must have the value of an **unsigned char** or EOF
```
emphasis mine.

You need to cast to `uint8_t` on each of these.

PR #389, opened 2020-07-10
Fix GCC warning

orgads:
```
include_server/c_extensions/distcc_pump_c_extensions_module.c: In function ‘XArgv’:
include_server/c_extensions/distcc_pump_c_extensions_module.c:214:13: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  214 |     argv[i] = PyUnicode_AsUTF8(string_object);    /* does not increase
      |             ^
```

Far from optimal, but fixing const-correctness will require many changes.


orgads:
@shawnl?

shawnl:
Oh yeah, and argv handling is already nutso as it is always pre-allocated.

PR #398, opened 2020-10-13
Fix compile errors while detecting C99 snprintf

### LINKED ISSUE #397: Build failure with Xcode 12 (clang 1200.0.32.2)
Building with Xcode 12 on macOS 10.15 leads to the following failure in `src/snprintf.c`:

```
2020-10-07T13:49:56.8826920Z src/snprintf.c:815:6: error: expected parameter declarator
2020-10-07T13:49:56.8827510Z  int vsnprintf (char *str, size_t count, const char *fmt, va_list args)
2020-10-07T13:49:56.8828190Z      ^
2020-10-07T13:49:56.8829450Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:75:40: note: expanded from macro 'vsnprintf'
2020-10-07T13:49:56.8830350Z   __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
2020-10-07T13:49:56.8830760Z                                        ^
2020-10-07T13:49:56.8831710Z src/snprintf.c:815:6: error: expected ')'
2020-10-07T13:49:56.8833330Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:75:40: note: expanded from macro 'vsnprintf'
2020-10-07T13:49:56.8834210Z   __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
2020-10-07T13:49:56.8834610Z                                        ^
2020-10-07T13:49:56.8835420Z src/snprintf.c:815:6: note: to match this '('
2020-10-07T13:49:56.8836710Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:75:29: note: expanded from macro 'vsnprintf'
2020-10-07T13:49:56.8838360Z   __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
2020-10-07T13:49:56.8839100Z                             ^
2020-10-07T13:49:56.8840890Z src/snprintf.c:815:6: error: conflicting types for '__builtin___vsnprintf_chk'
2020-10-07T13:49:56.8841960Z  int vsnprintf (char *str, size_t count, const char *fmt, va_list args)
2020-10-07T13:49:56.8842400Z      ^
2020-10-07T13:49:56.8843740Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:75:3: note: expanded from macro 'vsnprintf'
2020-10-07T13:49:56.8844620Z   __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
2020-10-07T13:49:56.8845010Z   ^
2020-10-07T13:49:56.8846220Z src/snprintf.c:815:6: note: '__builtin___vsnprintf_chk' is a builtin with type 'int (char *, unsigned long, int, unsigned long, const char *, struct __va_list_tag *)'
2020-10-07T13:49:56.8847860Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:75:3: note: expanded from macro 'vsnprintf'
2020-10-07T13:49:56.8848740Z   __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
2020-10-07T13:49:56.8849140Z   ^
2020-10-07T13:49:56.8850100Z src/snprintf.c:815:6: error: definition of builtin function '__builtin___vsnprintf_chk'
2020-10-07T13:49:56.8850790Z  int vsnprintf (char *str, size_t count, const char *fmt, va_list args)
2020-10-07T13:49:56.8851190Z      ^
2020-10-07T13:49:56.8852600Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:75:3: note: expanded from macro 'vsnprintf'
2020-10-07T13:49:56.8853480Z   __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
2020-10-07T13:49:56.8853870Z   ^
2020-10-07T13:49:56.8854740Z src/snprintf.c:817:29: error: use of undeclared identifier 'fmt'
2020-10-07T13:49:56.8855240Z     return dopr(str, count, fmt, args);
2020-10-07T13:49:56.8855550Z                             ^
2020-10-07T13:49:56.8856400Z src/snprintf.c:817:34: error: use of undeclared identifier 'args'
2020-10-07T13:49:56.8856930Z     return dopr(str, count, fmt, args);
2020-10-07T13:49:56.8857240Z                                  ^
2020-10-07T13:49:56.8857920Z src/snprintf.c:830:6: error: expected parameter declarator
2020-10-07T13:49:56.8858550Z  int snprintf(char *str,size_t count,const char *fmt,...)
2020-10-07T13:49:56.8858920Z      ^
2020-10-07T13:49:56.8860170Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:57:39: note: expanded from macro 'snprintf'
2020-10-07T13:49:56.8861030Z   __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
2020-10-07T13:49:56.8861390Z                                       ^
2020-10-07T13:49:56.8862440Z src/snprintf.c:830:6: error: expected ')'
2020-10-07T13:49:56.8863750Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:57:39: note: expanded from macro 'snprintf'
2020-10-07T13:49:56.8864600Z   __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
2020-10-07T13:49:56.8865510Z                                       ^
2020-10-07T13:49:56.8866440Z src/snprintf.c:830:6: note: to match this '('
2020-10-07T13:49:56.8867920Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:57:28: note: expanded from macro 'snprintf'
2020-10-07T13:49:56.8868770Z   __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
2020-10-07T13:49:56.8869150Z                            ^
2020-10-07T13:49:56.8870050Z src/snprintf.c:830:6: error: conflicting types for '__builtin___snprintf_chk'
2020-10-07T13:49:56.8870690Z  int snprintf(char *str,size_t count,const char *fmt,...)
2020-10-07T13:49:56.8871070Z      ^
2020-10-07T13:49:56.8872490Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:57:3: note: expanded from macro 'snprintf'
2020-10-07T13:49:56.8873340Z   __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
2020-10-07T13:49:56.8873720Z   ^
2020-10-07T13:49:56.8874860Z src/snprintf.c:830:6: note: '__builtin___snprintf_chk' is a builtin with type 'int (char *, unsigned long, int, unsigned long, const char *, ...)'
2020-10-07T13:49:56.8876440Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:57:3: note: expanded from macro 'snprintf'
2020-10-07T13:49:56.8877290Z   __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
2020-10-07T13:49:56.8877670Z   ^
2020-10-07T13:49:56.8878590Z src/snprintf.c:830:6: error: definition of builtin function '__builtin___snprintf_chk'
2020-10-07T13:49:56.8879260Z  int snprintf(char *str,size_t count,const char *fmt,...)
2020-10-07T13:49:56.8879640Z      ^
2020-10-07T13:49:56.8880930Z /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/secure/_stdio.h:57:3: note: expanded from macro 'snprintf'
2020-10-07T13:49:56.8882010Z   __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
2020-10-07T13:49:56.8882390Z   ^
2020-10-07T13:49:56.8883330Z src/snprintf.c:835:18: error: use of undeclared identifier 'fmt'
2020-10-07T13:49:56.8883800Z     va_start(ap, fmt);
2020-10-07T13:49:56.8884050Z                  ^
2020-10-07T13:49:56.8887190Z clang -DMINILZO_HAVE_CONFIG_H -DHAVE_CONFIG_H -D_GNU_SOURCE -DLIBDIR="\"/usr/local/Cellar/distcc/3.3.3_3/lib\"" -DSYSCONFDIR="\"/usr/local/Cellar/distcc/3.3.3_3/etc\"" -DPKGDATADIR="\"/usr/local/Cellar/distcc/3.3.3_3/share/distcc\"" -Isrc -I"./src" -I"./lzo" -I"./popt" -Werror -I/private/tmp/distcc-20201007-10856-1qqh2j4/distcc-3.3.3/include -MD -W -Wall -Wimplicit -Wshadow -Wpointer-arith -Wcast-align -Wwrite-strings -Waggregate-return -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Wmissing-declarations -D_THREAD_SAFE  -o src/loadfile.o -c src/loadfile.c
2020-10-07T13:49:56.8890020Z src/snprintf.c:836:33: error: use of undeclared identifier 'fmt'
2020-10-07T13:49:56.8890540Z     ret = vsnprintf(str, count, fmt, ap);
2020-10-07T13:49:56.8890860Z                                 ^
2020-10-07T13:49:56.8891140Z 12 errors generated.
2020-10-07T13:49:56.8891660Z make: *** [src/snprintf.o] Error 1
```

mitchblank:
This test produces errors on newer compilers (Xcode 12, probably others)  This means that it incorrectly fails, leading to
distcc trying to use its own `snprintf()`.  This then has compile issues with Xcode 12, breaking the build.

Fixes #397

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -433,9 +433,12 @@ fi
 
 AC_CACHE_CHECK([for C99 vsnprintf],rsync_cv_HAVE_C99_VSNPRINTF,[
 AC_TRY_RUN([
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 #include <sys/types.h>
 #include <stdarg.h>
-void foo(const char *format, ...) {
+static void foo(const char *format, ...) {
        va_list ap;
        int len;
        char buf[5];
@@ -449,7 +452,7 @@ void foo(const char *format, ...) {
 
        exit(0);
 }
-main() { foo("hello"); }
+int main() { foo("hello"); }
 ],
 rsync_cv_HAVE_C99_VSNPRINTF=yes,rsync_cv_HAVE_C99_VSNPRINTF=no,rsync_cv_HAVE_C99_VSNPRINTF=cross)])
 if test x"$rsync_cv_HAVE_C99_VSNPRINTF" = x"yes"; then
--- a/configure.ac
+++ b/configure.ac
@@ -62,14 +62,13 @@ getpeername(0,0,&len);
 dnl Android requires -llog for printf and friends
 dnl and will only run position independent executables
 AC_MSG_CHECKING([if building on Android])
-AC_TRY_COMPILE([
+AC_TRY_COMPILE([], [
   #ifdef __BIONIC__
   int ok;
   (void)ok;
   #else
   choke me
   #endif],
-[func("a"); func("a", "b"); func("a", "b", "c")],
 [AC_MSG_RESULT(yes)
  LDFLAGS="$LDFLAGS -llog -pie"],
 [AC_MSG_RESULT(no)])
@@ -399,12 +398,11 @@ if (mmap (NULL, 0, 0, 0, 0, 0) == MAP_FAILED)
 	return 0;
 #else
 #error mmap unavailable
-#endif], ,[
+#endif], AC_MSG_RESULT(yes), [
+	AC_MSG_RESULT(no)
 	AC_DEFINE([MAP_FAILED], [(void *)-1L],
 	[Define if MAP_FAILED constant not available])
 ])
-AC_MSG_RESULT()
-
 
 dnl
 dnl Test if the preprocessor understand vararg macros
@@ -438,7 +436,7 @@ AC_TRY_RUN([
 #include <string.h>
 #include <sys/types.h>
 #include <stdarg.h>
-static void foo(const char *format, ...) {
+static int foo(const char *format, ...) {
        va_list ap;
        int len;
        char buf[5];
@@ -446,13 +444,13 @@ static void foo(const char *format, ...) {
        va_start(ap, format);
        len = vsnprintf(0, 0, format, ap);
        va_end(ap);
-       if (len != 5) exit(1);
+       if (len != 5) return 1;
 
-       if (snprintf(buf, 3, "hello") != 5 || strcmp(buf, "he") != 0) exit(1);
+       if (snprintf(buf, 3, "hello") != 5 || strcmp(buf, "he") != 0) return 1;
 
-       exit(0);
+       return 0;
 }
-int main() { foo("hello"); }
+int main() { return foo("hello"); }
 ],
 rsync_cv_HAVE_C99_VSNPRINTF=yes,rsync_cv_HAVE_C99_VSNPRINTF=no,rsync_cv_HAVE_C99_VSNPRINTF=cross)])
 if test x"$rsync_cv_HAVE_C99_VSNPRINTF" = x"yes"; then
@@ -465,9 +463,9 @@ AC_TRY_RUN([
 #include <sys/types.h>
 #include <sys/socket.h>
 
-main() {
+int main() {
        int fd[2];
-       exit((socketpair(AF_UNIX, SOCK_STREAM, 0, fd) != -1) ? 0 : 1);
+       return (socketpair(AF_UNIX, SOCK_STREAM, 0, fd) != -1) ? 0 : 1;
 }],
 rsync_cv_HAVE_SOCKETPAIR=yes,rsync_cv_HAVE_SOCKETPAIR=no,rsync_cv_HAVE_SOCKETPAIR=cross)])
 if test x"$rsync_cv_HAVE_SOCKETPAIR" = x"yes"; then


mitchblank:
Actually, looking more at the configure.ac, there are a number of m4 mistakes that are breaking some of the tests.  Let me push a more comprehensive fix to my branch

configure.ac
```diff
@@ -62,14 +62,13 @@ getpeername(0,0,&len);
...
 AC_MSG_CHECKING([if building on Android])
-AC_TRY_COMPILE([
+AC_TRY_COMPILE([], [
   #ifdef __BIONIC__
   int ok;
   (void)ok;
   #else
   choke me
   #endif],
-[func("a"); func("a", "b"); func("a", "b", "c")],
```
mitchblank, line 72: This seems to have just been a copy/paste from another test.  I can't see how this test could have ever worked right, though, since the body of the `main()` needs to be the second arg to `AC_TRY_COMPILE()`.  The `(void)ok;` would have ended up outside of the function body even on Adroid!

configure.ac
```diff
@@ -399,12 +398,11 @@ if (mmap (NULL, 0, 0, 0, 0, 0) == MAP_FAILED)
 	return 0;
 #else
 #error mmap unavailable
-#endif], ,[
+#endif], AC_MSG_RESULT(yes), [
+	AC_MSG_RESULT(no)
 	AC_DEFINE([MAP_FAILED], [(void *)-1L],
 	[Define if MAP_FAILED constant not available])
 ])
-AC_MSG_RESULT()
```
mitchblank, line 406: This was just cosmetic, but the `./confugure` run wasn't correctly printing whether this test passed or not.

shawnl:
Maybe C11 changed it and I didn't know, but implicit `int` is very much a C thing and the code looks correct (if annoying). I think XCode is just being pedantic and is the broken thing (which wouldn't suprise me given they they refuse to install C headers by default these days).

configure.ac
```diff
@@ -433,9 +433,12 @@ fi
 
 AC_CACHE_CHECK([for C99 vsnprintf],rsync_cv_HAVE_C99_VSNPRINTF,[
 AC_TRY_RUN([
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 #include <sys/types.h>
 #include <stdarg.h>
-void foo(const char *format, ...) {
+static void foo(const char *format, ...) {
```
shawnl: This change especially looks wrong. While supporting old compilers is not really important, `static` is a new thing and completely unnecessary here.

mitchblank:
>  think XCode is just being pedantic and is the broken thing

Well, it was an intentional change on their part - see https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes :
> Clang now reports an error when you use a function without an explicit declaration when building C or Objective-C code for macOS (-Werror=implicit-function-declaration flag is on). This additional error detection unifies Clang’s behavior for iOS/tvOS and macOS 64-bit targets for this diagnostic. (49917738)

configure.ac
mitchblank: It isn't 100% needed to fix the problem, but I was just trying to make the code fragment compile warning-free to make it as reliable as possible.  Is there any reason that you think it shouldn't be marked `static`?  It clearly isn't an exported function.

shawnl:
Yeah, Apple being broken solely for the sake of being broken. I am not going to bite.

**macports is still supported.** (but I do not have this platform and cannot test)

No interest in supporting purposefully broken things on broken platforms.

The horrible truth about Apple's repeated engineering failures - Louis Rossman https://www.youtube.com/watch?v=AUaJ8pDlxi8

1:01 - A1226/A1260 2007-2008 Macbook GPU failures, warranty service refusal
2:21 - A1226/A1260 2007-2008 Macbook Pro hinge/frame problem
3:16 - A1286 Macbook Pro - the "Unibody" myth, glued together pieces fall apart
4:58 - A1286/A1297 MCP power circuit failure due to poor buck converter design: C7771 issue
6:01 - iPhone 4 cellular placement fail
7:12 - iPhone 5 power button problem
7:27 - A1286 2010 Macbook Pro GPU kernel panics due to same buck converter defect from 2008/2009(this gives you a hint that apple engineers doesn't give a crap about engineering good products, same design flaw for three straight years)
10:04 - A1286 2011 Macbook Pro GPU failure, Apple gets sued over not addressing problem. 
11:43 - Apple gives out badly refurbished boards as warranty replacements for 2011 GPU failures.
13:06 - 2012 Retina Macbook Pro: another motherboard issue (U8900), due to poor soldering/manufacturing method on the GPU buck converter. 
14:46 - Mac Pro GPU failure (again).
16:27 - iPhone 6/6+ touchscreen issue due to structural issue. 
18:23 - SSD soldered straight into the motherboard+ chip that would kill the macbook, because a power line would short out to ground when the chip dies.
20:18 - 2016 Macbook keyboard reliability issue.
21:52 - 2016 Macbook Battery failure issue.
22:50 - A1278 Macbook Pro SATA cable failures(yes, really).

shawnl:
If you can fix this error with a simple "is this macOS" + adding a flag, then I will take it, but changing the tests (and introducing `static` flag which defeats the whole purpose of autotools) from what is totally standard C is not acceptable.

configure.ac
shawnl: because `static` is a C99 keyword, and it makes the test no long C89. While we don't support C89 that defeats the whole purpose of what autotools is.

shawnl:
I am sorry if this seems bitter, but breaking standard-compliant autotools tests breaks the majority of free software, so they are doing this just to attract attention, and it is really lame.

shawnl:
I did have a patch moving to meson-build, which would deal with this in a more correct way (while also dropping a number of platforms).

shawnl:
I think the correct way to fix this is to test for macOS, and then disable that error. Having that flag that way on the code is fine, but the test is correct, and does not need to be changed.

configure.ac
mitchblank: That is just not true.  `static` existed long before C99, or even C89.  It was in use even in the original Bell Labs compilers in the 1970s.

Here's an example from Dennis Richie himself: https://github.com/dspinellis/unix-history-repo/blob/Research-V7-Snapshot-Development/usr/src/libc/gen/ttyslot.c#L40

It predates the `<stdarg.h>` header by at least a decade, so there is no risk that a compiler would be old enough to pass this test and not support something as basic as `static`.  Indeed, a UNIX machine that old wouldn't even be able to run `./configure` since the Bourne shell wouldn't be featureful enough.

mitchblank:
> I think the correct way to fix this is to test for macOS, and then disable that error. Having that flag that way on the distcc code is fine, but the test is correct, and does not need to be changed.

I find that opinion pretty bizarre.

A well written autoconf test will have a nice clean code fragment that just verifies the functionality that it's looking for.  The code in this test was very much *not* clean -- it compiled with a bunch of warnings even when it worked.  For example, it was missing needed `#include` statements so if a function that it was using was actually implemented using a macro, the compile would have failed.

My change just makes the test code compile cleanly, which should make the test more reliable overall.  I don't see any downside to doing this in general, plus it fixes the problem at hand.

Your take seems to be that this autoconf test was *intentionally* written to be fragile -- so rather than just cleaning up the test, you want a platform-specific workaround.  I honestly can't comprehend where you are coming from with this.

configure.ac
```diff
@@ -438,21 +436,21 @@ AC_TRY_RUN([
 #include <string.h>
 #include <sys/types.h>
 #include <stdarg.h>
-static void foo(const char *format, ...) {
+static int foo(const char *format, ...) {
```
shawnl, line 439: Ok I see that it already uses static.

shawnl:
OK, I looked closer and implicit `int` was removed in C99.

PR #399, opened 2020-10-13
clirpc.c -- fix unreachable code

mitchblank:
Fix bug indicated by clang warning:
```
src/clirpc.c:126:9: warning: code will never be executed [-Wunreachable-code]
```
It looks like ~12 years ago someone added a log statement to this `if()` block while they had Python on their mind.  This means that the "expected version" check has been broken that entire time!

### DIFF
--- a/src/clirpc.c
+++ b/src/clirpc.c
@@ -114,7 +114,7 @@ int dcc_r_result_header(int ifd,
     unsigned vers;
     int ret;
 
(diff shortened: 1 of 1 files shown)


shawnl:
Thanks!

Yes, this means there hasn't been any version mis-matches all this time. (and I am one of the few people that actually knows the protocol enough to know what that means)

PR #407, opened 2021-01-03
Port distccmon-gnome from gtk2 to gtk3

nickoe:
GTK 2 has reached it's end of life: https://blog.gtk.org/2020/12/16/gtk-4-0/

So I thought this would be a good exercise for me to port this as I use distccmon-gnome from time to time and would like not to loose it. 

This should be ready to go and compiles without warnings for me.

Attached is a screenshot of both:
Selection_195

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,10 +132,10 @@ AC_ARG_ENABLE(Werror,
 # we need.  It's faster to just do it once during configuration.
 if test "x${with_gnome}" = xyes
 then
-    GNOME_PACKAGES="gtk+-2.0 libgnome-2.0 libgnomeui-2.0 pango"
+    GNOME_PACKAGES="gtk+-3.0 libgnome-3.0 libgnomeui-3.0 pango"
 elif test "x${with_gtk}" = xyes
 then
-    GNOME_PACKAGES="gtk+-2.0"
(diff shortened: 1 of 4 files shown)


shawnl:
Wow, this is amazing!

I am having a minor issue where the title is half-size, both in 200% scaling mode (I have a 4K screen), and in 100% scaling mode.

Captura de pantalla de 2021-01-04 21-20-51

nickoe:
@shawnl Does other gtk3 apps use "proper" scaling for the window title?I don't own any hdpi monitor. There are no deprecation warnings shown, even though I enabled that option locally. How does the gtk2 version look on your hdpi monitor?

I added this before configure:
```
export CFLAGS="-DGDK_DISABLE_DEPRECATED -DGTK_DISABLE_DEPRECATED -DGTK_DISABLE_SINGLE_INCLUDES"
```
And it builds without warnings after my patches.

shawnl:
I am very familiar with the core distcc code, but not pump mode or gtk. This looks good, and that is a minor bug. You can however use screen scaling without having a hidpi monitor, in your monitor settings, and set 200%.

nickoe:
@shawnl Ok, thank you very much for merging, but I don't use the same window manager (I use i3) as you so I can't see the titlebar scaling easily.

PR #411, opened 2021-01-05
Remove debug print that is no longer needed

nickoe:
@shawnl Woops, there was a small debug print that I forgot to remove in my local rebase :S

Please review and merge this before tagging a new release :D

### DIFF
--- a/src/mon-gnome.c
+++ b/src/mon-gnome.c
@@ -630,15 +630,7 @@ int main(int argc, char **argv)
    * compilation */
   nice(5);
 
-#if defined(WITH_GNOME)
-  gnome_program_init ("distccmon-gnome", PACKAGE_VERSION,
-              LIBGNOMEUI_MODULE,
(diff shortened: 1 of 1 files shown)


shawnl:
Yes, I build with --with-gtk and not --with-gnome too, and @thedrow also tried to contradict me when I suggested that the Debian package make the same switch, and I didn't do this myself because I didn't care to contradict the distcc author.

And a printf() instead of a dcc log call is definitely a bug.

Thanks!

thedrow:
I wanted to be absolutely sure we compile everywhere since I'm not an expert on the code base.
I am certainly not the author of distcc. I just made sure it was salvaged from Google code.
If #gtk says it's fine, it is. 😊

shawnl:
Might have not been you, it was a *long* time ago on Debian bug tracker, and might have even been before gtk 3 release. :)

PR #413, opened 2021-02-08
Skip distributing LTO cc invocations

RX14:
When distcc is configured as the system compiler, sometimes it ends up being used on LTO builds. In this case it's faster to run the `cc` invocations locally, since most of the work of the compiler is done in the linker.

I'm guessing they'll be some discussion on this feature but I thought I'd rather post a patch.

### DIFF
--- a/src/arg.c
+++ b/src/arg.c
@@ -189,6 +189,9 @@ int dcc_scan_args(char *argv[], char **input_file, char **output_file,
                 rs_trace("-mtune=native optimizes for local machine; "
                          "must be local");
                 return EXIT_DISTCC_FAILED;
+            } else if (!strcmp(a, "-flto")) {
+                rs_trace("LTO cc invocations are not worth distributing");
+                return EXIT_DISTCC_FAILED;
             } else if (str_startswith("-Wa,", a)) {
                 /* Look for assembler options that would produce output
                  * files and must be local.


shawnl:
Makes sense to me.

shawnl:
I need to remember to include this in release notes however

shawnl:
Wait.....did you test that LTO builds even work in the first place? I dont think they would build at all.

shawnl:
@RX14 ^^

RX14:
They work, just locally. This patch makes distcc pass through the command to `cc` verbatim, (just like `-march=native`) so distcc acts like it's not there.

It actually worked before this patch too, it sent the `cc` invocations to the remote host and did the actual opt and codegen in `ld` on the host. The only thing that got distributed was the parsing passes, and it's not worth it. Hence this patch.

AnonymousRetard:
I am running 3 Gentoo systems with distcc configured as the system compiler. I have also globally enabled LTO flags for all packages with a few exceptions which dont work with LTO. I have been wondering for a long time why very, very few packages seem to build faster with DISTCC enabled while most actually seem to build a bit slower and don't really seem to be distributing work very well.

Building the Linux kernel seems to be one of the few things that do distribute well. (Which is one of the few things I don't have LTO enabled on).

I guess this explains it. I never realized it's not possible to distribute packages with -flto using DISTCC. I guess DISTCC doesn't really make sense with my setup :-/.

I can definitely confirm that -flto builds over DISTCC without issues, but it doesn't really speed anything up, it seems to actually slow building down by a little bit. Now I will have to consider removing LTO flags on my slower machines if I want to be able to distribute and speed up their compiling.

asheplyakov:
> and did the actual opt and codegen in ld on the host.

That's not quite correct. GCC always generates the (target) machine code for every compilation unit. With `-flto` GCC also saves the internal (GIMPLE)  representation of the code (into a special section of the object file).

> In this case it's faster to run the cc invocations locally

That depends on lot of factors, including, but not limited to:

1. The number of local cores
2. The number of remote machines/cores
3. How fast the local and remotes cores are
4. If the local machine has enough RAM (especially for compiling C++)

I believe not distributing LTO compilations unconditionally is definitely wrong.

shawnl:
@asheplyakov 

I didn't put much thought into this before merging it. There are also things like ThinLTO, and I am not really up on everything.

What behavior do you think distcc should have? What about throwing an error when called with `-flto`?

Issue #416, opened 2021-02-24
cross-compilation with clang fails


asheplyakov:
```
$ cat > foo.c <<-EOF
int main(int argc, char **argv) { return 0; }
EOF
$ env DISTCC_HOSTS=127.0.0.1 distcc clang --target=aarch64-linux-gnu- -c foo.c
error: unknown target triple 'unknown', please use -triple or -arch
distcc[304936] ERROR: compile foo.c on 127.0.0.1 failed
distcc[304936] (dcc_build_somewhere) Warning: remote compilation of 'foo.c' failed, retrying locally
distcc[304936] Warning: failed to distribute foo.c to 127.0.0.1, running locally instead
error: unknown target triple 'unknown', please use -triple or -arch
distcc[304936] ERROR: compile foo.c on localhost failed
```

Linux' kernel makefiles specify target arch with `--target=foo` rather than `-target foo`, hence cross-compilation with distcc and clang fails.

PR #417, opened 2021-02-24
Improved cross-compilation with clang

asheplyakov:
With this patch I can cross-compile Linux kernel with clang and distcc.
Otherwise both remote compilation and local fallback fail like this:

```
cat > foo.c <<-EOF
int main(int argc, char **argv) { return 0; }
EOF
env DISTCC_HOSTS=127.0.0.1 distcc clang -c --target=aarch64-linux-gnu- foo.c

error: unknown target triple 'unknown', please use -triple or -arch
distcc[304773] ERROR: compile foo.c on 127.0.0.1 failed
distcc[304773] (dcc_build_somewhere) Warning: remote compilation of 'foo.c' failed, retrying locally
distcc[304773] Warning: failed to distribute foo.c to 127.0.0.1, running locally instead
error: unknown target triple 'unknown', please use -triple or -arch
distcc[304773] ERROR: compile foo.c on localhost failed
```

Linux' kernel makefiles specify the target arch with `--target=foo`,
and distcc expects `-target foo`, hence the problem.

Closes: #416

### DIFF
--- a/src/argutil.c
+++ b/src/argutil.c
@@ -67,6 +67,19 @@ int dcc_argv_search(char **a,
 }
 
 
+/**
+ * Return true if argv contains argument starting with needle.
+ */
+int dcc_argv_startswith(char **a,
+                        const char *needle)
+{
+    size_t needle_len = strlen(needle);
+    for (; *a; a++)
+        if (!strncmp(*a, needle, needle_len))
+            return 1;
+    return 0;
+}
+
 unsigned int dcc_argv_len(char **a)
 {
     unsigned int i;
--- a/src/compile.c
+++ b/src/compile.c
@@ -557,7 +557,12 @@ static void dcc_add_clang_target(char **argv)
     else
         return;
 
+    /* -target aarch64-linux-gnu */
     if (dcc_argv_search(argv, "-target"))
+        return;
+
+    /* --target=aarch64-linux-gnu */
+    if (dcc_argv_startswith(argv, "--target"))
         return;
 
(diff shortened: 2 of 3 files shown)


src/argutil.c
```diff
@@ -67,6 +67,19 @@ int dcc_argv_search(char **a,
...
 
 
+/**
+ * Return true if argv contains argument starting with needle.
+ */
+int dcc_argv_startswith(char **a,
+                        const char *needle)
+{
+    size_t needle_len = strlen(needle);
+    for (; *a; a++)
```
shawnl, line 77: This is unnecessary. `strncmp` does this for you.

shawnl (changes requested):
the string processing routine is broken

src/argutil.c
asheplyakov: Let's check it:

```c
char *args[] = { "clang", "-c", "--target=aarch64-linux-gnu", "foo.c", NULL };
int idx = 2;
if (dcc_argv_startswith(args, "--target=") != 1) {
    fprintf(stderr, "wrong answer: args[%d] ('%s') does contain '--target'\n", idx, args[idx]);
    exit(1);
}
```

Removing `for` loop from `distcc_argv_startswith` breaks the test.
Perhaps I'm missing something. Could you elaborate, please? What exactly strncmp is supposed to do?

shawnl: oh my bad, I didn't see the double star.

Issue #420, opened 2021-04-27
distccd is unable to make use of available cores/CPU time


asheplyakov:
When compiling a "big" C project (Linux kernel) compilation nodes never use 100% of CPU time (although there are a plenty of compilation tasks available). Instead a system is 20 -- 30% idle (despite --jobs is 3x the number of the cores).

PR #421, opened 2021-04-27
prefork: use available cores more efficiently

asheplyakov:
Observed behavior
-----------------

When compiling a big C project (Linux kernel) with distcc compilation nodes
never use 100% of CPU time, instead a system is 25 -- 40% idle (despite --jobs
is set to 3x the number of cores).

Analysis
--------

Linux kernel consists of many relatively small C sources and header files.
Compiling a single source file typically takes a fraction of a second (although
there are quite a number of "heavier" sources).
`dcc_preforked_child` exits after 50 requests. Thus a typical lifetime of
a worker process is just a few seconds. However `dcc_create_kids` sleeps for
a second after every fork. Also `dcc_preforked_child` sleeps for yet another second.
Which is >= 25% of a typical child lifetime. As a result distccd is unable to use
available CPU time (system load is always <= 75% during the compilation).

Fix
---

Removed `sleep` from `dcc_preforking_parent` and `dcc_create_kids`.
Also made `dcc_preforked_child` process compilation jobs for at least
60 seconds.

Result
------

Kernel compilation time (two 6-core nodes) reduced almost 2x:
before: 20 minutes, after: 12 minutes

Closes: #420

### DIFF
--- a/src/prefork.c
+++ b/src/prefork.c
@@ -37,6 +37,7 @@
 #include <syslog.h>
 #include <signal.h>
 #include <fcntl.h>
+#include <time.h>
 
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


shawnl:
I like you write-up. I will give this a shot.

Issue #426, opened 2021-07-11
distcc fails to find the native compiler 


asheplyakov:
```
cat > hello.c <<-EOF
#include <stdio.h>
int main(int argc, char **argv) {
     printf("Hello, world!\n");
     return 0;
}
EOF
distcc gcc -Wall -std=gnu89 -I. -O2 -o hello.o -c hello.c
```

The output is:
```
free(): invalid next size (fast)
Aborted (core dumped)
```

There are two unrelated problems here:
1. distcc (or rather ` dcc_gcc_rewrite_fqn`) fails to find the native compiler due to autoconf idea about canonical system name (`x86_64-alt-linux-gnu`) being different from that of GCC (`x86_64-alt-linux`)
2. `dcc_gcc_rewrite_fqn` writes beyond the allocated memory (`newcmd`), see #425 for more details)

Also problem 1 occurs if the toolchain uses a slightly different triple (say, `aarch64-unknown-linux-gnu` instead of the expected `aarch64-linux-gnu`)

asheplyakov:
> distcc (or rather dcc_gcc_rewrite_fqn) fails to find the native compiler due to autoconf idea about canonical system name (x86_64-alt-linux-gnu) being different from that of GCC (x86_64-alt-linux)

The same problem exists on Fedora (CentOS, RHEL):  gcc triplet is `x86_64-redhat-linux`, and configure identifies the system as `x86_64-pc-linux-gnu`

shawnl:
This is all really ugly. I know that I am the one that added this cross-compile stuff, but GCC is really at fault here. When I met Andrew Pinsky the inferiorness in cross-compile compared to clang was admitted, but no-body has come forward with a canonicalization, and done the work.

I am not sure what distcc can do, as we are kind of at the butt end of the problem.

asheplyakov:
> This is all really ugly. 

I agree.

> but GCC is really at fault here

Firstly those wonderful `${arch}-redhat-linux` triplets are sort of set in the stone (they have existed for more than 2 decades), so now it does not matter who/what is at fault.
Secondly just because GCC triplets are configurable doesn't mean it's GCC fault.

> I am not sure what distcc can do

What about guessing the correct triplet for distro/OS with `gcc -dumpmachine` (which is what this patch does)?

asheplyakov:
> What about guessing the correct triplet for distro/OS with gcc -dumpmachine (which is what this patch does)?

`this patch` is #427

[...]

Issue #412, opened 2021-02-01
DISTCC_NO_CROSS_REWRITE not working?


solskogen:
Either I don't understand how DISTCC_NO_CROSS_REWRITE should work, or perhaps DISTCC_NO_CROSS_REWRITE is broken on Debian. 
I've got a setup where my fast x86_64 machine is supposed to compile for my rather slow Raspberry Pi. On the x86_64 box I've got the cross compiler installed (aarch64-linux-gnu-gcc and aarch64-linux-gnu-g++). 
On the Pi I've got /usr/lib/distcc first in my PATH so `which gcc` gives me /usr/lib/distcc/gcc.  On the fast box I've only got aarch64-linux-gnu-gcc and aarch64-linux-gnu-g++ in /usr/lib/distcc, both which are symlinks which points to /usr/lib/distcc. 

However, when I compile on the Pi, the compilation fails due to the fact that gcc/g++ doesn't exist in /usr/lib/distcc on the fast box.  If I understand the manual correctly distcc should've rewritten them as aarch64-linux-gnu-gcc/aarch64-linux-gnu-g++ on the fast box. I've tried NOT having DISTCC_NO_CROSS_REWRITE defined, I've tried it set to 1, and even 0. Still no dice. 

If I on the Pi set CC to "distcc aarch64-linux-gnu-gcc" I can compile without any trouble.

wangpenghit2155:
correct answer is:  DISTCC_NO_REWRITE_CROSS 
you have mistaken it for DISTCC_NO_CROSS_REWRITE

solskogen:
I don't understand, could you explain it a bit better? The man page doesn't say anything about  DISTCC_NO_REWRITE_CROSS

shawnl:
That feature predates my cross-compiler rewrite logic, wherein you provide explicit remote compiler paths to use. I never used it.

asheplyakov:
This issue might be a duplicate of #440

@solskogen: Could you please post the verbose distcc log?
```
env DISTCC_VERBOSE=1 distcc gcc YOUR_COMMAND_LINE_HERE
```

solskogen:
pi@raspberrypi:~ $ env DISTCC_VERBOSE=1 distcc gcc -c test.c
distcc[1002] (dcc_trace_version) distcc 3.3.2 ; built May 12 2019 07:48:36
distcc[1002] (dcc_recursion_safeguard) safeguard level=0
distcc[1002] (main) compiler name is "distcc"
[...]

PR #427, opened 2021-07-11
Improved native compiler triple detection

asheplyakov:
Unfortunately the configure script and the compiler don't agree on canonical system name.
I.e. configure thinks it's `x86_64-pc-linux-gnu`, and GCC uses `x86_64-linux-gnu` instead.
Therefore ask the compiler (when running distcc configure script) how to correctly call it.

Closes: #426
ALTBUG: 40425


asheplyakov:
A gentle ping to see if I can get this merged.
Apparently people have been experiencing the problem for quite a long time (see #412)

Issue #431, opened 2021-07-21
Error: (dcc_check_compiler_whitelist) CRITICAL! no /usr/local/lib/distc


t0ny-peng:
This error pops up on the server side(the one that runs `distccd`. I have two machines, running Ubuntu 18.04. `distcc` is compiled from source:
```shell
cd /tmp && wget https://github.com/distcc/distcc/releases/download/v3.4/distcc-3.4.tar.gz
tar xf ./distcc-3.4.tar.gz
cd distcc-3.4
./configure --without-libiberty
make all -j$(nproc)
make install;
```
In the `make install` step  it prints the locations of the files to be installed
```
'make install' will install distcc as follows:
  man pages            /usr/local/share/man/man1
  documents            /usr/local/share/doc/distcc
  programs             /usr/local/bin
  sbin programs        /usr/local/sbin
  system configuration /usr/local/etc
  icon file            /usr/local/share/pixmaps
  application file     /usr/local/share/applications
```
Then to setup the masquade mode, I manually ran `sudo update-distcc-symlinks`.
The daemon is started by `
distccd --daemon --allow 10.0.0.0/8 --log-file=/tmp/distccd.log --verbose`. 

On the client side, `distcc` was installed using the same way. When calling `cmake ..` I do see that `distcc` is being used as the C++ compiler
```
-- Check for working CXX compiler: /usr/lib/distcc/bin/c++
-- Check for working CXX compiler: /usr/lib/distcc/bin/c++ - works
```
... but the compilation failed with message:
```
[  3%] Building C object 3rdparty/ippiw/CMakeFiles/ippiw.dir/src/iw_image_filter_box.c.o
distcc[12865] (dcc_build_somewhere) Warning: failed to distribute, running locally instead
```

And on the server side, the log file says:
```
distccd[3240] (dcc_r_file) received 253 bytes to file /tmp/distccd_77dbace9.ii
distccd[3240] (dcc_r_file_timed) 253 bytes received in 0.000322s, rate 767kB/s
distccd[3240] (dcc_set_input) changed input from "/tmp/opencv/build/CMakeFiles/CMakeTmp/testCXXCompiler.cxx" to "/tmp/distccd_77dbace9.ii"
distccd[3240] (dcc_set_input) command after: x86_64-linux-gnu-g++ -fPIE -o CMakeFiles/cmTC_2b53a.dir/testCXXCompiler.cxx.o -c /tmp/distccd_77dbace9.ii
[...]

PR #433, opened 2021-07-21
distccd: keep looking for approved compilers in /usr/lib/distcc ...

asheplyakov:
... instead of bailing out if `$prefix/lib/distcc` directory does not exist

Closes: #431


thedrow (approved):

shawnl:
LGTM

Issue #435, opened 2021-07-26
update-distcc-symlinks fails to find GCC compilers on non-x86 64-bit rpm-based systems


asheplyakov:
On ARM64 box running ALT Linux:
```
# update-distcc-symlinks
cc
gcc
clang
clang++
```

1. Versioned symlinks (`gcc-8`, `gcc-10`) are missing
2. `aarch64-alt-linux-gcc-8`, `aarch64-alt-linux-gcc-10` symlinks are missing

On x86_64 `update-distcc-symlinks` works correctly since native compilers being multilib (with both 32-bit and 64-bit runtimes, and 32-bit one resides in `/usr/lib`)

PR #436, opened 2021-07-26
masquerade mode improvements for rpm-based systems

asheplyakov:
On ARM64 rpm-based distros GCC bits are installed into `/usr/lib64/gcc` and `/usr/libexec/gcc` directories. However `update-distcc-symlinks` does not scan these directories. As a result `update-distcc-symlinks` does NOT create symlinks to
 
* versioned GCC binaries (`gcc-8`, `gcc-10`, etc)
* fully qualified GCC binaries (`aarch64-linux-gnu-gcc-8`, etc)

With this patchset `update-distcc-symlinks` correctly work on such systems.

Also distcc/distccd always use `/usr/lib/distcc` (both on 32-bit and 64-bit hosts) so the masquerade mode always works as advertised in the manual:
```
PATH=/usr/lib/distcc:$PATH make -j32
```

Closes: #435

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -59,7 +59,7 @@ include_server_builddir = $(builddir)/_include_server
 
 # These must be done from here, not from autoconf, because they can 
 # contain variable expansions written in Make syntax.  Ew.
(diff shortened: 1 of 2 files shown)


shawnl:
I wish we had some tests.....

Issue #434, opened 2021-07-25
DISTCC_SSH regression in v3.4


KSmanis:
Version 3.4 seems to be affected by a regression regarding `DISTCC_SSH`, according to which the environment variable is only honored during the first connection attempt of distcc and ignored during any subsequent runs (assuming the distcc server is temporarily unavailable).

I accidentally stumbled upon this issue when my CI started failing once Gentoo stabilized the relevant version.

Relevant CI log (truncated):
<details>

```
distcc-ssh-client_1  | distcc[10] (dcc_trace_version) distcc 3.4 i686-pc-linux-gnu; built Jul  4 2021 03:18:57
distcc-ssh-client_1  | distcc[10] (dcc_recursion_safeguard) safeguard level=0
distcc-ssh-client_1  | distcc[10] (main) compiler name is "distcc"
distcc-ssh-client_1  | distcc[10] (dcc_scan_args) scanning arguments: gcc -c test.c -o test-distcc.o
distcc-ssh-client_1  | distcc[10] (dcc_scan_args) found input file "test.c"
distcc-ssh-client_1  | distcc[10] (dcc_scan_args) found object/output file "test-distcc.o"
distcc-ssh-client_1  | distcc[10] compile from test.c to test-distcc.o
distcc-ssh-client_1  | distcc[10] (dcc_get_hostlist) read hosts from environment
distcc-ssh-client_1  | distcc[10] (dcc_parse_hosts) found ssh token "distcc-ssh@distcc-ssh-server"
distcc-ssh-client_1  | distcc[10] (dcc_lock_host) got cpu lock on distcc-ssh@distcc-ssh-server slot 0 as fd3
distcc-ssh-client_1  | distcc[10] (dcc_lock_host) got cpu lock on localhost slot 0 as fd4
distcc-ssh-client_1  | distcc[10] (dcc_note_state) note state 3, file "(NULL)", host "localhost"
distcc-ssh-client_1  | distcc[10] (dcc_strip_dasho) result: gcc -c test.c
distcc-ssh-client_1  | distcc[10] (dcc_spawn_child) forking to execute: gcc -E test.c
distcc-ssh-client_1  | distcc[10] (dcc_spawn_child) child started as pid11
distcc-ssh-client_1  | distcc[10] (dcc_strip_local_args) result: gcc -c test.c -o test-distcc.o
[...]

PR #438, opened 2021-08-30
Restrict number of retries when backoff is disabled

asheplyakov:
Since the commit a7dd5cf90e8c ("Try a new host after failing to connect instead of immediately falling back to localhost") distcc tries to choose a different host on connection/protocol errors. However when backoff is disabled (DISTCC_BACKOFF_PERIOD=0) distcc keeps retrying forever.
    
To avoid the problem limit the number of retries (currently to 3) when backoff is disabled.

Closes: #434

### DIFF
--- a/src/backoff.c
+++ b/src/backoff.c
@@ -52,6 +52,19 @@
 
 static int dcc_backoff_period = 60; /* seconds */
 
+static int dcc_get_backoff_period(void)
+{
+    char *bp;
+    bp = getenv("DISTCC_BACKOFF_PERIOD");
+    if (bp)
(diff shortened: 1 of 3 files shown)


shawnl:
Why not just revert a7dd5cf ?

asheplyakov:
> Why not just revert [a7dd5cf](https://github.com/distcc/distcc/commit/a7dd5cf90e8c38e9ec00ca00dfc5aaa3a0ee138d) ?

Because trying a different remote node is a more sensible behavior and it works well, except in some corner cases (i.e. disabled backoff + only one remote node)

PR #439, opened 2021-09-05
distccd: add --oom-score-adj option (Linux only)

whitslack:
This new option sets `/proc/self/oom_score_adj` in prefork worker child processes. This can afford users a bit of protection against a scenario in which many concurrent compilation jobs cause an out-of-memory condition that prompts the kernel's OOM killer to kill a process that is not so "disposable" as a distccd worker.

This commit adds a new Autoconf test to check whether the host system type (that is, the type of the system on which the compiled binaries are to be run) is Linux. The new `--oom-score-adj` option is implemented only if the host system type is Linux.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,6 +15,18 @@ AC_INIT(distcc, 3.4, distcc@lists.samba.org)
 AC_CONFIG_HEADERS(src/config.h)
 AC_CONFIG_MACRO_DIRS([m4])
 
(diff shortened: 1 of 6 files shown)


shawnl:
Sounds like a good idea.

Does this inherit to the compiler?

whitslack:
> Does this inherit to the compiler?

Yes. The score is inherited across `fork` and preserved across `exec`.

Issue #443, opened 2021-11-05
Server rejects IPv4 clients for no good reason


asheplyakov:
Built with
```   
./configure --enable-rfc2553 && make && make install
```    

Started server:
```    
distccd --listen :: --allow-private --log-level=debug --log-file=/tmp/distccd.log
```

Clients connecting via IPv4 get (erroneously) rejected:
```
distccd[10488] (dcc_check_client) connection from ::ffff:10.64.MM.NN:36744
distccd[10488] (check_address_inet6) deny v6 client
distccd[10488] (check_address_inet6) deny v6 client
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0x7f, mask 0xff
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0x10ac, mask 0xf0ff
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0xa, mask 0xff
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0xa8c0, mask 0xffff
```

PR #444, opened 2021-11-05
Fixed access control with IPv4-mapped addresses

asheplyakov:
`inet_from_inet6` returns a little endian representation of the address, which breaks the access control for IPv4 clients:

```
distccd[10488] (dcc_check_client) connection from ::ffff:10.64.MM.NN:36744
distccd[10488] (check_address_inet6) deny v6 client
distccd[10488] (check_address_inet6) deny v6 client
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0x7f, mask 0xff
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0x10ac, mask 0xf0ff
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0xa, mask 0xff
distccd[10488] (check_address_inet) deny client 0xa40MMNN, value 0xa8c0, mask 0xffff
```

As a result clients connecting via IPv4 get (erroneously) rejected.
Use a simple(r) conversion method to avoid the problem.

Closes: #443

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -316,17 +316,9 @@ static int check_address_inet6(const struct in6_addr *client,
 /**
  * Extract the lower 32-bits of a v6 address.
  **/
-static uint32_t inet_from_inet6(const struct in6_addr *addr)
+static void inet_from_inet6(struct in_addr *v4addr, const struct in6_addr *addr)
 {
-    uint32_t dest = 0;
-    int i;
-
-    for (i = 0; i < 4; ++i) {
-        uint32_t dest_byte = addr->s6_addr[12 + i];
-        dest |= (dest_byte << (8 * (3-i)));
-    }
-
-    return dest;
+    memcpy(&v4addr->s_addr, addr->s6_addr + 12, 4);
 }
 #endif /* ENABLE_RFC2553 */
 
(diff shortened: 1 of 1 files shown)


asheplyakov:
A gentle ping to see if I can get this merged.
#443 makes using distcc in a mixed IPv4/IPv6 environment tricky.

martinetd (approved):
For what it's worth, I can confirm the problem exists and that this fixes it for me.

src/access.c
```diff
@@ -361,7 +353,7 @@ int dcc_check_address(const struct sockaddr *client,
         if (value->family == AF_INET) {
             if (IN6_IS_ADDR_V4MAPPED(a6) || IN6_IS_ADDR_V4COMPAT(a6)) {
                 struct in_addr a4;
-                a4.s_addr = inet_from_inet6(a6);
+                inet_from_inet6(&a4, a6);
```
sourcefrog, line 356: This seems to leave the rest of the in_addr uninitialized, which might not matter in this case but seems like a bad idea. Could you zero it or set the fields you know?

Otherwise this looks good, thanks!

sourcefrog (changes requested):

src/access.c
sourcefrog: Actually I'll do that and then merge. Thanks for the patch!

asheplyakov: > This seems to leave the rest of the in_addr uninitialized, which might not matter in this case but seems like a bad idea. Could you zero it or set the fields you know?

Done

sourcefrog (approved):

PR #451, opened 2022-01-06
daemon: --allow-private fixes

martinetd:
The three commits are widely independent, if you don't agree with any of them I'd suggest just dropping the dubious patches, merge what we can, and continue with the other(s) more leisurely.

- running a standalone daemon with `--allow-private` would fail on alpine, because rfc2553 is not explicitly enabled and the code chokes on ipv6 address parsing.
- running a standalone daemon with no `--allow` option would print a big warning that `--allow-private` is assumed instead, but not actually do anything with it
- lastly I'd suggest enabling rfc2553 by default, I'm not sure why it's not. If there's a compatibility problem somewhere make it autodetected? But I haven't found one and would assume it's more widely used? Looking at debian/fedora, debian uses it but not fedora...

Thanks!


akarle:
@martinetd -- a *huge* thanks for catching this! I dropped the configure defaults change because I'm not sure how packagers will feel / what the impact will be, but the --allow-private fixes feel like a big security lapse (assuring users it's restricted when it's not).

I just force-pushed and will merge when CI passes--I added a bit more detail to the commit messages tracking down the original PR / regression point (#240, released in 3.3). I'm hoping to get a release going soon, and this'll definitely warrant the cut.

The only functional change I added to your code was removing the external int from dopt.h--opt_allow_private is no longer used anywhere else in the codebase so it can be local again.

P.S. as an extra bonus, it looks like the old code was missing the last element of the IP range anyways (`<6` instead of `<=6`); another win :)

akarle:
Hm @sourcefrog it looks like I can't merge without a second approver:

>     New changes require approval from someone other than akarle because they were the last pusher.

It's unclear if it's because I force-pushed to update the branch or if I'm the last pusher on master...

This one definitely warrants merging / a new release / callout in the changelog IMHO.

martinetd:
Thanks! I personally think the default change is fine nowadays, but I agree it's worth handling separately from the fixes.

I've checked your rebase and just repushed the same content if that helps :)

PR #459, opened 2022-04-29
Support `-f*-prefix-map` compiler options

rmatev:
In pump mode, distcc rewrites the debug info on the server before
returning the result (see `dcc_fix_debug_info`). However, when
`-ffile-prefix-map` or similar compiler options are passed,
the compiler is unable to perform the substitution since the path on
the server is different. The rewriting of distcc still kicks in,
and the net result on the client side is that the `-prefix-map`
option is not respected.

This is fixed by simply treating the `-f*-prefix-map` options
analogously to the include arguments.

### DIFF
--- a/src/serve.c
+++ b/src/serve.c
@@ -485,6 +485,49 @@ static int tweak_input_argument_for_server(char **argv,
 }
 
 
+static const char *prefix_map_options[] = {
+    "-ffile-prefix-map=",
+    "-fmacro-prefix-map=",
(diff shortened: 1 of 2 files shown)


rmatev:
this is really tested with #460 , but the two PRs can go in independently.

mgulick:
My company has been applying this patch to a custom build of distcc, as we make use of `-fdebug-prefix-map`.  Is there anything I can do to help get this upstreamed?

PR #468, opened 2022-08-21
distccd: set autogroup niceness (Linux only)

whitslack:
As described in ["Why nice levels are a placebo and have been for a very long time, and no one seems to have noticed"][1], the Linux kernel in its default configuration on many Linux distributions autogroups tasks by session ID and "fairly" allocates CPU time among such autogroups. The nice levels of tasks within each autogroup are relevant only relative to other tasks within the same autogroup. Effectively, this means that the traditional nice level is rendered moot when a daemon runs in its own session and thus in its own autogroup. Linux does provide a means to change the niceness of autogroups relative to each other, so let's have distccd make use of this feature where available so that `-N`,`--nice` will actually do what the user intends. On systems where autogroups are not supported or are disabled, this commit introduces no change in behavior.

[1]: https://www.reddit.com/r/linux/comments/d7hx2c/why_nice_levels_are_a_placebo_and_have_been_for_a/

### DIFF
--- a/src/dparent.c
+++ b/src/dparent.c
@@ -334,6 +334,42 @@ void dcc_remove_pid(void)
 
 
 /**
+ * Sets the niceness of this process's autogroup, if applicable.
+ *
+ * When Linux is running with autogroups enabled, each session gets allocated
+ * its own autogroup, and the task scheduler tries to allocate CPU time
+ * "fairly" among all such autogroups. The nicenesses of the processes within
+ * an autogroup is relevant only versus other processes within the same
+ * autogroup. Thus, when a process is the only task in its autogroup, or when
+ * all tasks in its autogroup have the same niceness, then the process's
+ * individual niceness is utterly irrelevant. To have the intended niceness
+ * apply relative to other autogroups, the niceness of the autogroup itself
+ * must be set. This function does that.
+ **/
+static void dcc_set_autogroup_niceness(int niceness)
+{
+#ifdef HAVE_LINUX
+    FILE *fp = fopen("/proc/self/autogroup", "r+");
+    int error = 0;
+
+    if (fp) {
+        if (fprintf(fp, "%d\n", niceness) < 0) {
+            error = errno;
+        } else {
+            rs_trace("autogroup niceness: %d", niceness);
+        }
+        if (fclose(fp) == EOF && !error)
+            error = errno;
+    } else if (errno != ENOENT) {
+        error = errno;
+    }
+    if (error)
+        rs_log_error("autogroup nice %d failed: %s", niceness, strerror(error));
+#endif
+}
+
+
+/**
  * Become a daemon, discarding the controlling terminal.
  *
  * Borrowed from rsync.
@@ -367,6 +403,7 @@ static void dcc_detach(void)
         rs_log_error("setsid failed: %s", strerror(errno));
     } else {
         rs_trace("setsid to session %d", (int) sid);
+        dcc_set_autogroup_niceness(opt_niceness);
     }
 #else /* no HAVE_SETSID */
 #ifdef TIOCNOTTY


sourcefrog:
I wonder if the simplest thing is to just not use the option?

Modern Linux generally seems to do pretty well at balancing even very cpu-hungry parallel builds with interactive tasks, with no need for manual nice levels.

Also, the daemon is presumably most often run on dedicated build servers, where there is no need for it to be nice to other tasks?

whitslack:
@sourcefrog: That's an odd response. Don't use the option? I'm sharing this improvement specifically because I *do* want to use the option, and I want it to work, which it currently does not. I'm not concerned about "balancing even very cpu-hungry parallel builds with interactive tasks"; I'm concerned about balancing remote parallel builds versus local parallel builds versus non-interactive but high-priority services. Yes, I do run all of that on one machine, and I'd like to make use of niceness to prioritize my limited CPU cycles.

sourcefrog:
Out of curiosity, does it have a measurable effect? You didn't mention. 

If you have a machine that's running multiple non interactive workloads at different priorities then turning off auto groups seems like a better first step? 

Anyhow, I am no longer maintaining distcc, so active maintainers can decide if they think it's a good idea to merge.

whitslack:
> Out of curiosity, does it have a measurable effect? You didn't mention.

**Test setup:** Quad-core desktop system running `distccd` and four instances of `setsid xz </dev/urandom >/dev/null 2>&1` to ensure that all the CPU cores are reliably loaded with work. Owing to `setsid`, each `xz` process runs in its own session and thus in its own autogroup.

**Test scenario:** Compiling SQLite on a Raspberry Pi configured to perform all compilations solely on the aforementioned desktop system via distcc (i.e., no local compilations).

**Control case:** `distccd` niceness set to 9; its autogroup niceness set to 0 (the default). `distccd` logs a time of 4m 15s to compile SQLite's monolithic translation unit.

**Experimental case:** `distccd` niceness set to 9; its autogroup niceness set (manually) to 19. `distccd` logs a time of 59m 5s to compile SQLite's monolithic translation unit.

@sourcefrog: So yes, significant, measurable effect. This is as expected. If there were no practical effect, it would indicate a severe bug in the kernel's task scheduler.

> If you have a machine that's running multiple non interactive workloads at different priorities then turning off auto groups seems like a better first step?

Turning off autogroups is not viable. I have a service that runs thousands of subprocesses concurrently, and I want the entire service as a whole to be treated fairly by the scheduler relative to other services. Turning off autogrouping would give this massively multi-process service a huge, unfair advantage over the other, single-process services running on the same box. For that matter, distccd is a multi-process service, and I want the kernel's task scheduler to treat it as a single unit for purposes of fair scheduling, regardless of how many distcc jobs happen to be running at any given instant.

sourcefrog:
Interesting. I would not have imagined wanting to run distccd on a machine (or container) that's also 100% occupied by several other distinct workloads. 

An hour to compile a single large file sounds barely usable, but I suppose if it's in batch perhaps it still is useful.

whitslack:
@sourcefrog: It's not 100% occupied typically. That's why I contrived some `xz` instances to demonstrate that autogroup niceness matters. I want distcc jobs not to slow down anything else on the machine at all, but I do still expect them to suck up any cycles that would otherwise be idle. See?

sourcefrog:
Yes, I see. Is it impractical to set this on whatever creates the session/group containing distccd? I guess that is what you are doing in the experiment?

sourcefrog:
Yes, I see. Is it impractical to set this on whatever creates the session/group containing distccd? I guess that is what you are doing in the experiment? 

If I was writing this now I probably would not have added `--nice`, because it can be reasonably set externally.

whitslack:
@sourcefrog: Well, the problem with setting it externally is that it has to be done *after* `distccd` is running: use `pidof` to find the PID of `distccd` (any of the matching processes will do), and then write the new autogroup niceness to `/proc/<pid>/autogroup`. (Yes, that's what I was doing for my experiment.) It can't be done *prior* to exec'ing `distccd` because `distccd` (rightly) calls `setsid` to create a new session, and the new session gets put into a new autogroup that the kernel creates with default niceness. In other words, autogroup niceness is not inherited across a `setsid` call, so niceness must be set on the new autogroup **after** the `setsid` call, and the most convenient place from which to do that is inside the daemon itself.

sourcefrog:
Ah, wow, that's awkward. Ok.

whitslack:
Indeed! Not only is it awkward, but it actually seems like it could be a gross oversight in the kernel design, as it allows an unprivileged, deprioritized process to elevate its scheduling priority simply by calling `setsid` to place itself into a new autogroup with default niceness. If I had been designing autogroups, I would have made autogroup niceness inherited across `setsid` to preclude such priority escalation attacks.

asheplyakov:
What about using systemd facilities to control resource limits (instead of re-implementing the same logic in distcc)?

whitslack:
> What about using systemd facilities to control resource limits (instead of re-implementing the same logic in distcc)?

@asheplyakov: Not everyone uses systemd. I sure don't. Regardless, though, this is about autogroups, and — correct me if I'm wrong — autogroups are not hierarchical, so it wouldn't matter what systemd does with the autogroup niceness on the parent process because distccd will call `setsid` to start a new session, which the kernel will place into a virgin autogroup that doesn't inherit its parent's niceness.

asheplyakov:
> > What about using systemd facilities to control resource limits (instead of re-implementing the same logic in distcc)?
> 
> @asheplyakov: Not everyone uses systemd. I sure don't. 

Use your favorite service manager to configure resource limits for distccd. Otherwise you end up implementing a (poor) service manager within distccd (which does not sound like a bright idea).

> Regardless, though, this is about autogroups, and — correct me if I'm wrong — autogroups are not hierarchical

Yes, this is correct.

> so it wouldn't matter what systemd does with the autogroup niceness on the parent process because distccd will call `setsid` to start a new session,

distcc does NOT call `setsid` when `--no-detach` is specified.

whitslack:
> distccd does NOT call `setsid` when `--no-detach` is specified.

@asheplyakov: Hmm, and Gentoo's OpenRC runscript for distccd _does_ specify `--no-detach`, and my `distccd` processes do in fact have their autogroup niceness set to 0 despite that I have distccd installed with this patch applied and am launching it with `-N 9`. So all this time my patch has been doing nothing for me! I wonder why I never noticed. Regardless, thank you for pointing this out.

> Use your favorite service manager to configure resource limits for distccd. Otherwise you end up implementing a (poor) service manager within distccd (which does not sound like a bright idea).

I do agree with that reasoning. However, doesn't that same argument also apply against distccd's existing `--nice` option? The service manager should be setting the niceness before exec'ing `distccd --no-detach`. Given that you do offer an option to set the niceness, you should really make it do what it says on the label, which is what this PR accomplishes.

Now, with this patch actually being irrelevant in a `--no-detach` setup, I need to fix my service configuration so that my `distccd` processes are _actually_ deprioritized as I intended them to be. Thanks again for the tip-off.

asheplyakov:
> However, doesn't that same argument also apply against distccd's existing `--nice` option?

Actually it does apply, and I would remove `--nice` option (or make it a no-op to not break existing setups).

asheplyakov:
> distccd does NOT call `setsid` when `--no-detach` is specified.

However it does call `setpgid` (even when NOT detaching) [before running the compiler](https://github.com/distcc/distcc/blob/master/src/exec.c#L402). The compiler typically spawns a lot of processes (`cc1`, `cc1plus`, assemblers, whatever). The idea is to kill these processes [if the client disconnects too early](https://github.com/distcc/distcc/blob/master/src/exec.c#L522). However the kernel autogrouping **might** create a new control group for the compiler (and make nice level useless).

whitslack:
> > distccd does NOT call `setsid` when `--no-detach` is specified.
> 
> However it does call `setpgid` […] the kernel autogrouping **might** create a new control group for the compiler (and make nice level useless).

Currently, Linux only creates an autogroup per session, not per process group, so calling `setpgid()` is not enough to cause a new autogroup to be created. (If it were, then every process you might start from an interactive shell command line would run in its own autogroup, which would kind of defeat the purpose of autogroups.) The intention behind autogroups is to prevent one abusive user from monopolizing CPU resources by forking a large number of tasks. In my opinion, autogroups really should have been designed around UID rather than SID, but per-session autogroups is what we have today.

PR #469, opened 2022-09-03
migrate CI from Travis to GitHub Actions

andy5995:


andy5995:
The results can be viewed at https://github.com/andy5995/distcc/actions

Issue #454, opened 2022-01-17
Python3.10 pump error


himekifee:
Trying to compile something with pump mode on. Pump exit with stderr shown below
```
WARNING include server: Preprocessing locally. Include server internal error: '<class 'SystemError'>: ("PY_SSIZE_T_CLEAN macro must be defined for '#' formats",)' for translation unit '/home/xxx/xxx/xxx.c'
------- Include server stack trace -----------
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/include_server/include_server.py", line 384, in handle
    DoCompilationCommand(cmd, currdir,
  File "/usr/lib/python3.10/site-packages/include_server/include_analyzer.py", line 261, in DoCompilationCommand
    self.ProcessCompilationCommand(currdir, parsed_command))
  File "/usr/lib/python3.10/site-packages/include_server/include_analyzer.py", line 189, in ProcessCompilationCommand
    total_closure.update(self._ProcessFileFromCommandLine(translation_unit,
  File "/usr/lib/python3.10/site-packages/include_server/include_analyzer.py", line 129, in _ProcessFileFromCommandLine
    fpath_resolved_pair, fpath_real = self.build_stat_cache.Resolve(
  File "/usr/lib/python3.10/site-packages/include_server/cache_basics.py", line 779, in Resolve
    if _OsPathIsFile(relpath):
SystemError: PY_SSIZE_T_CLEAN macro must be defined for '#' formats
----------------------------------------------
```

Muha0644:
I am also having the same issue. 
From what I've seen it's something that python 3.10 broke

TornaxO7:
I have the same issue as well.

Issue #474, opened 2022-10-16
Include Server Not covering


ElectricPulse:
I am using pump to compile the linux kernel and I get these types of errors on the client:
`Include server not covering: Could not find include file 'include/linux/... for translation unit. arch/x86/kernel...`

The server log looks fine.
The non pump mode works - but it doesnt utilize system recources very well and I suspect its due to the heavy preprocessing of the linux kernel compilation.

ElectricPulse:
To add some extra information:
The error is from a ubuntu system
The include path of include_server.py is correct
This happens on my self built version and on the one from apt
This also happens on my arch system aswell

manorit2001:
Facing similar issue. Ubuntu 22.04 with apt installation of distcc
```
WARNING include server: Preprocessing locally. Include server not covering: Could not find include file 'include/linux/kconfig.h'. for translation unit 'drivers/media/mc/mc-dev-allocator.c'
```

PR #479, opened 2023-01-15
Replace int with Py_ssize_t for distcc pump extension

hephooey:
Defining PY_SSIZE_T_CLEAN is required since python 3.10, and the types of length variables should be modified to match it. Otherwise functions like OsPathExists seem always return False in my tests.

The PY_SSIZE_T_CLEAN macro is supported at least back to python 3.5 according to docs.python.org, so I included it without any python version conditions.

This PR should fix https://github.com/distcc/distcc/issues/454 and likely https://github.com/distcc/distcc/issues/474 as well

### DIFF
--- a/include_server/c_extensions/distcc_pump_c_extensions_module.c
+++ b/include_server/c_extensions/distcc_pump_c_extensions_module.c
@@ -21,6 +21,7 @@
 /* distcc_pump_c_extensions_module.c -- Python bindings for distcc-pump
  * extensions */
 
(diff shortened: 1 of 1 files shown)


z-aki:
@shawnl  requesting a review here

mattst88:
Looks like @shawnl has no github activity in more than a year. @sourcefrog has merged the last few pull requests.

sourcefrog:
This looks reasonable to me, so I'll merge it, but I don't think I want to become the main maintainer again.

Issue #497, opened 2024-01-14
"Segmentation fault'" when run distcc with undefined args or PATH followed string "distcc"


GeorgeFrancisco:
Hi,
In dcc_which(), the 'while ((path = strchr(path, ':') + 1))' will always be ture because of '+ 1', it will cause 'Segmentation fault' in: 

1.when host removed gcc and run distcc with any undefined args, for example:
#apt remove gcc
#distcc -v
Segmentation fault
#distcc -h
Segmentation fault

2.when PATH environment variables followed a path string includes 'distcc', for example:
#which gcc
/usr/bin/gcc
#echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:**/root/distcc**
#distcc -v
Segmentation fault

Version: b88432459aff2fb240a138b5fe8150db67ca6eb3(Latset)
Reproduce rate: 100%

Please find complete cmdline to reproduce this issue in attached cmdline.txt.
[cmdline.txt](https://github.com/distcc/distcc/files/13930914/cmdline.txt)

Please find verbose log in attached distcc.log.
[distcc.log](https://github.com/distcc/distcc/files/13930915/distcc.log)

PR #498, opened 2024-01-14
Fix issue #497 "Segmentation fault'" when run distcc with undefined …

GeorgeFrancisco:
…args or PATH followed string "distcc"

In dcc_which(), the 'while((path = ... + 1))' will always be ture because of '+ 1', it will cause 'Segmentation fault' in: 1.when host removed gcc and run 'distcc -v' or 'distcc -h' or any other undefined args. 2.when PATH environment variables followed a path string includes 'distcc'.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
 
 int dcc_which(const char *command, char **out)
 {
-    char *loc = NULL, *_loc, *path, *t;
-    int ret;
+    char *loc = NULL, *path, *t;
+    int len, ret;
 
     path = getenv("PATH");
     if (!path)
         return -ENOENT;
+
     do {
-        if (strstr(path, "distcc"))
-            continue;
-        /* emulate strchrnul() */
         t = strchr(path, ':');
         if (!t)
-            t = path + strlen(path);
-        _loc = realloc(loc, t - path + 1 + strlen(command) + 1);
-        if (!_loc) {
-            free(loc);
-            return -ENOMEM;
+           len = strlen(path);
+        else
+            len = t - path;
+        loc = realloc(loc, len + strlen(command) + 2);
+        if (!loc) {
+             free(loc);
+             return -ENOMEM;
         }
-        loc = _loc;
-        strncpy(loc, path, t - path);
-        loc[t - path] = '\0';
+        strncpy(loc, path, len);
+        loc[len] = '\0';
         strcat(loc, "/");
         strcat(loc, command);
         ret = access(loc, X_OK);
-        if (ret < 0)
+        if (strstr(loc, "distcc") || ret < 0) {
+            path = t + 1;
             continue;
+        }
         *out = loc;
         return 0;
-    } while ((path = strchr(path, ':') + 1));
+    } while(t);
     return -ENOENT;
 }
 


sourcefrog:
Thanks! I wish we had C unit tests for this kind of thing, which would make it easier to confirm what was fixed, but I guess I didn't think that way at the time, and maybe it's too much to add now...

src/util.c
```diff
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
...
-        _loc = realloc(loc, t - path + 1 + strlen(command) + 1);
-        if (!_loc) {
-            free(loc);
-            return -ENOMEM;
+           len = strlen(path);
+        else
+            len = t - path;
+        loc = realloc(loc, len + strlen(command) + 2);
+        if (!loc) {
+             free(loc);
```
sourcefrog, line 600: It's not a huge deal because this will only happen on ENOMEM, which should be rare, but: the `free` seems pointless, since you know `loc` is null here. You could keep the old one in another variable, as the old code does.

src/util.c
```diff
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
 
 int dcc_which(const char *command, char **out)
```
sourcefrog, line 583: Can you add a comment about the purpose of this function, which is essentially: 

> search for _command_ in every colon-separated directory inside _PATH_...

except with some special behavior if the generated path contains "distcc"? 

I don't know or remember what that's for -- git blame might show you -- but I guess it's about avoiding loops when distcc itself is on the path?

src/util.c
```diff
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
...
     path = getenv("PATH");
     if (!path)
         return -ENOENT;
+
     do {
-        if (strstr(path, "distcc"))
-            continue;
-        /* emulate strchrnul() */
         t = strchr(path, ':');
         if (!t)
```
sourcefrog, line 594: I realize the names are already here but I think they'd be clearer as, say renaming `t` to `next_colon`, and `len` to `dir_len`...

src/util.c
```diff
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
...
-        loc = _loc;
-        strncpy(loc, path, t - path);
-        loc[t - path] = '\0';
+        strncpy(loc, path, len);
+        loc[len] = '\0';
         strcat(loc, "/");
         strcat(loc, command);
         ret = access(loc, X_OK);
-        if (ret < 0)
+        if (strstr(loc, "distcc") || ret < 0) {
```
sourcefrog, line 608: The behavior is a bit different here: the old code checks whether the directory name contains `distcc` but here you're checking whether the full generated path contains `distcc` and, I'm not sure if that's intentional?

src/util.c
```diff
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
...
+            path = t + 1;
```
sourcefrog, line 609: If we're in the last component of the path, then `t` (or let's call it `next_colon`) will be NULL, and so this sets it to `(char *) 1`, doesn't it? And it seems like then it might crash on the next iteration? At least, the `while` condition won't see it as 0.

src/util.c
```diff
@@ -582,35 +582,36 @@ int dcc_remove_if_exists(const char *fname)
...
             continue;
+        }
         *out = loc;
         return 0;
-    } while ((path = strchr(path, ':') + 1));
+    } while(t);
```
sourcefrog, line 614: The code seems a little complex here with a `continue` and `return` right before the end of the loop.

Maybe we could simplify it to:

1. If the directory name contains "distcc" then skip it before building the whole path or calling `access`
2. If `access` found it, then save the path and return
3. If `next_colon` is not null (there are more paths) then step past the colon and continue
4. Otherwise, return ENOENT

What do you think?

sourcefrog (changes requested):
Thank you!

PR #500, opened 2024-01-21
Fix issue distcc#497 "Segmentation fault'" when run distcc with undef…

GeorgeFrancisco:
…ined args or PATH followed string "distcc"

In dcc_which(), the 'while((path = ... + 1))' will always be ture because of '+ 1', it will cause 'Segmentation fault' in: 1.when host removed gcc and run 'distcc -v' or 'distcc -h' or any other undefined args. 2.when PATH environment variables followed a path string includes 'distcc'.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -580,37 +580,49 @@ int dcc_remove_if_exists(const char *fname)
     return 0;
 }
 
+/* 
+ * Returns the absolute path of *command by searching each subpath of PATH.
+ * 
+ * Input *command is either "c++" or "cc".
+ * **out points to the absolute path, such as "/usr/bin/cc".
+ * 
+ * Skip the path include "distcc" to avoid loops.
+ */
 int dcc_which(const char *command, char **out)
 {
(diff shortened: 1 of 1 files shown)


GeorgeFrancisco:
I deleted my fork make the old pull request canceled...

So here is a new one.

src/util.c
```diff
@@ -580,37 +580,49 @@ int dcc_remove_if_exists(const char *fname)
...
+        strncpy(loc, path, dir_len);
+        loc[dir_len] = '\0';
+        if (strstr(loc, "distcc"))
+            continue;
         strcat(loc, "/");
         strcat(loc, command);
         ret = access(loc, X_OK);
-        if (ret < 0)
+        if (ret < 0) {
+            path = next_colon + 1;
```
sourcefrog, line 620: This seems to still have the bug that path ends up set to 1?

If you iterate on this please keep the same branch and PR and push new commits, so that the comment history is not lost.

sourcefrog:
I merged #511 which builds on this. Thanks!

PR #506, opened 2024-03-06
Port CI to GitHub Actions

sourcefrog:
distcc has config to build on travis-ci.org, but that seems to be turned down (https://www.travis-ci.com/blog/2021-05-07-orgshutdown/). We could migrate to their travis-ci.com but GitHub's native CI probably gives an easier experience, and I'm more familiar with it.

Previously: #222 suggests building on more different platforms, but we could look at that separately; this shouldn't block it if we can find runners.

This doesn't try to build on Windows. Apparently autoconf used to work there, or can work there (under Cygwin?) but it looks nontrivial enough that I'll leave it for later.


sourcefrog:
cc @asheplyakov in case you have an opinion

PR #523, opened 2024-06-10
feat(stats): [Linux] Report `dcc_free_mem` from `/proc/meminfo`

### LINKED ISSUE #521: Feature request: Return free/available/allocateable RAM via the `--stats` output
Hey!

I am working on a [client-side load-balancing script](http://github.com/whisperity/distcc-driver/tree/feat/v2-implementation) for my team, as heavy-weight C++ developers we have been avid `distcc` users for quite some time now. I am and we are really happy to see the `--stats` output becoming a thing with `distcc` v*3*. Parsing and understanding **`dcc_max_kids`** is really great at lifting the need from the clients to configure how much jobs the worker servers can consume (this is an ever-changing value for us, which is why I'm making the a client-side automation do this sort of balancing) because we observed running too many local jobs or failing too many remote jobs resulting in retries will kill the developer machine due to technicalities outside our direct control.

Suppose a server in `DISTCC_HOSTS` can run **`16`** parallel jobs. There is a big difference in running 16× jobs which require let's say **2 GiB** of RAM each, versus running 16× **8 GiB** jobs. If I have an estimate on how big on average a job requires from the compiler (this estimate I can obtain, or at least have my clients reasonably guess...), then I could set `DISTCC_HOSTS` to distribute only **`4`** jobs for compilations that use more RAM per source file. But doing this division _all the time_ is not feasible, because another machine could exist that has enough RAM to deal with a larger number of larger jobs. Moreover, RAM use can fluctuate during the day based on how much memory load (and **not** CPU load, which `--stats` already reports) the machines receive.

Unfortunately, **`dcc_free_space`** is not a terribly useful metric, because it only reports the size of the mount point under `/tmp` (which is easy to know and control to a fixed value, especially if the servers are run containerised [[1](http://github.com/whisperity/distcc-docker), [2](http://github.com/kwk/distcc-docker-images), [3](http://github.com/choldrim/docker-distcc), [4](http://github.com/bensuperpc/docker-distcc), &hellip;]). The size of the object file can often exhibit no linear correlation with the actual total/peak memory use of the compiler that produced it.

----

**Proposal:** Add an extra entry to the `--stats` output response, e.g., _`dcc_free_mem`_, which reports the amount of available memory that could be allocated by the jobs being sent to the server, as returned in the last column of `free -m`.

This way, a load balancing script could query the available memory size and do a division and distribute a smaller number of jobs to a server if running the build of a project where it is known in advance that the average memory consumption of the compiler will be higher per-process.

whisperity:
Closes #521.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -627,10 +627,72 @@ int dcc_which(const char *command, char **out)
     }
 }
 
-/* Returns the number of processes in state D, the max non-cc/c++ RSS in kb and
- * the max RSS program's name */
-void dcc_get_proc_stats(int *num_D, int *max_RSS, char **max_RSS_name) {
-#if defined(linux)
+#ifdef linux
+/* Returns the value from the "MemAvailable" line in the opened file stream,
+ * that is expected to have the /proc/meminfo (proc(5)) format, converted to MB
+ * (or -1 if the information is not available).
+ * The file is closed by this function! */
+static int dcc_get_proc_meminfo_mem_available(FILE* f) {
+    int mem_available = -1;
+    char line[512];
+
+    char key[128];
+    long value;
+    char unit[4];
+    char magnitude;
+
+    if (f == NULL)
+        return -1;
+
+    while (fgets(line, sizeof(line), f) != NULL) {
+        if (sscanf(line, "%127s %ld %4s", key, &value, unit) != 3)
+            continue;
+        if (strncmp(key, "MemAvailable:", sizeof("MemAvailable:")) == 0)
+            break;
+    }
+
+    if (feof(f) || ferror(f)) {
+        fclose(f);
+        return -1;
+    }
+    fclose(f);
+
+    mem_available = value;
+    magnitude = unit[0];
+
+    switch (magnitude) {
+        case 'T':
+            mem_available *= 1024;
+            /* fallthrough */
+        case 'G':
+            mem_available *= 1024;
+            /* fallthrough */
+        case 'M':
+            break;
+        case 'K':
+        case 'k':
+            mem_available /= 1024;
+            break;
+        default:
+            rs_log_warning("/proc/meminfo: 'MemAvailable' is \"%d %.4s\", "
+                           "unhandled unit '%c'!",
+                           mem_available, unit, magnitude);
+            mem_available = -1;
+            break;
+    }
+
+    return mem_available;
+}
+#endif
+
+/* Returns the number of processes in state D, the max non-cc/c++ RSS in kB,
+ * the max RSS program's name, and the amount of available memory reported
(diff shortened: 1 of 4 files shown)


src/util.c
```diff
@@ -627,9 +627,67 @@ int dcc_which(const char *command, char **out)
     }
 }
 
-/* Returns the number of processes in state D, the max non-cc/c++ RSS in kb and
- * the max RSS program's name */
-void dcc_get_proc_stats(int *num_D, int *max_RSS, char **max_RSS_name) {
+/* Returns the value from the "MemAvailable" line in the opened file, converted
```
sourcefrog, line 630: I guess this will compile anywhere but please wrap it in `#ifdef linux` for clarity?

src/util.c
```diff
@@ -627,9 +627,67 @@ int dcc_which(const char *command, char **out)
...
+ * to MB. The file is closed by this function! */
```
sourcefrog, line 631:
```suggestion
 * to MB, or -1 on error. The file is closed by this function! */
```

src/util.c
sourcefrog, line 630:
```suggestion
/* Returns the value from the "MemAvailable" line in the opened Linux /proc/meminfo file, converted
```

src/util.c
```diff
@@ -656,11 +715,20 @@ void dcc_get_proc_stats(int *num_D, int *max_RSS, char **max_RSS_name) {
...
     *num_D = 0;
+    *mem_available = -1;
     *max_RSS = 0;
     *max_RSS_name = RSS_name;
     RSS_name[0] = 0;
 
     while ((procsubdir = readdir(proc)) != NULL) {
+        if (!checked_meminfo &&
+                strncmp(procsubdir->d_name, "meminfo", sizeof("meminfo"))) {
+            f = fopen("/proc/meminfo", "r");
```
sourcefrog, line 726: Not a big deal, but I think this doesn't really need to be inside the loop? You could just attempt to read it before or after the loop, and if for some reason it's not readable you'll get -1.

src/util.c
```diff
@@ -627,9 +627,67 @@ int dcc_which(const char *command, char **out)
...
+            mem_available = -1;
+            break;
+    }
+
+    return mem_available;
+}
+
+/* Returns the number of processes in state D, the max non-cc/c++ RSS in kB,
+ * the max RSS program's name, and the amount of available memory reported
+ * by the kernel. */
```
sourcefrog, line 686:
```suggestion
 * by the kernel (in MB, or -1 if unknown). */
```

sourcefrog (commented):
Looks good! Please also add it as an unreleased change in NEWS. 

The macOS CI breakage seems unrelated.

sourcefrog (approved):

PR #531, opened 2024-10-11
Add sysroot option and allow -specs option from compiler

wzssyqa:
Currently we reject the `-specs` option from compiler. Maybe it is due to we cannot be sure that the specs file is correct.

To support it, we can add a --sysroot option for the server, so that the server will try to look for the specs file from SYSROOT.


wzssyqa:
The test failure on macOS is not my fault, right?

sourcefrog:
> The test failure on macOS is not my fault, right?

Yep, it's pre-existing

sourcefrog:
This looks reasonable but I'd like to first try to understand more about why it was blocked in the first place. 

Thanks for adding tests!

PR #542, opened 2025-02-21
Fix a compile warning

black-desk:
> ```
> src/util.c:648:56: error: 'sscanf' may overflow; destination buffer
> in argument 5 has size 4, but the corresponding specifier may require
> size 5 [-Werror,-Wfortify-source]
>
>   648 |         if (sscanf(line, "%127s %ld %4s", key, &value, unit) != 3)
>       |                                                        ^
> 1 error generated.
> ```

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -638,7 +638,7 @@ static int dcc_get_proc_meminfo_mem_available(FILE* f) {
 
     char key[128];
     long value;
-    char unit[4];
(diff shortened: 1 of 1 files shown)


akarle:
FYI I force pushed to your branch to pick up the latest changes (which should kick off CI, needed for merge). I also amended the commit message to have a bit more detail

Issue #71, opened 2015-06-06
cross-compilation with +zeroconf


GoogleCodeExporter:
```
The zeroconf has hardcoded "cc -dumpmachine" and "cc -dumpversion" to broadcast 
compiler version available/required. The makes it very inconvenient for 
cross-compiler users.

BUG:
The distcc should use the gcc passed in, not the hardcoded one while searching 
for a matching distccd.

FEATURE REQUEST:
Ideally distccd should broadcast list of available compilers or at least allow 
for selecting non-default one and use multiple distccd instances. 


```

Original issue reported on code.google.com by `tomek.b...@gmail.com` on 15 Oct 2010 at 4:35

PR #544, opened 2025-02-21
Add option to ignore compiler version when using zeroconf

### LINKED ISSUE #541: Adding `+zeroconf` results in errors
I am trying to run `distcc` using zeroconf. I do have two `distccd` instances in my network. 

But when I add `+zeroconf` to the `~/distcc/hosts` file or to `DISTCC_HOSTS` I get the following error:
```
distcc[178988] (dcc_parse_hosts) Warning: /home/raskolnikov/.distcc/zeroconf/hosts contained no hosts; can't distribute work
distcc[178988] (dcc_zeroconf_add_hosts) CRITICAL! failed to parse host file.

distcc[178988] (dcc_show_hosts) CRITICAL! Failed to get host list
```

arximboldi:
By default, `distcc` uses `cc -version` to determine whether a host found via zeroconf is "compatible".

However, there are lots of use-cases where `cc` is ignored by the build process. Different machines may have different `cc`, yet provide compatible builds because the builds specify the compiler in a more explicit way.  See for example:

  https://github.com/distcc/distcc/issues/541
  https://github.com/distcc/distcc/issues/71

This adds a special token `+zeroconf.all` that is like `+zeroconf` but ignores `cc` when finding machines.

Closes #541


akarle (commented):
Thanks for taking the time to put this together.

Can you tell me a little more about your Avahi/zeroconf setup?

I'm leaning towards these changes being OK because it's net-new/backwards compatible, but I wonder if we might want to retain _some_ level of filtering. IIUC, it will search for this tag:

```
(daemon_proc) Browsing for '_distcc._tcp'.
```

I buy that `cc -version` isn't the most useful bit of info, but wouldn't it be useful to at least have the machine (`x86_64-linux-gnu`) in there? You wouldn't want to farm to different architecture hosts... right? Or is there some other filtering mechanism I'm unaware of? (this is my first look at avahi)

PR #545, opened 2025-03-01
Add initial CMake support

Ryanf55:
# Background

Hello! I just started using distcc and saw some of the maintainers complaining about autotools and others requesting CMake support. I had a few hours to kill, so here's a stab at adding modern (CMake 3.23+) support to distcc. 

So far, I've gotten `distcc.c` to compile on Ubuntu 22 with cmake 3.31 (I use latest released CMake for development).

# Demo

```
cd distcc
cmake -B build
cmake --build build
./build/src/client --help
```
Results in:
```
distcc[86199] ERROR: failed to exec client: No such file or directory
distcc[86198] ERROR: compile (null) on localhost failed with exit code 110
```

It compiles and doesn't segfault, so I thought I'd share my results now.

# Technical approach

In case you aren't that familar with CMake, here's the current high level approach:
* Use CMake's pkgconfig integration to find `popt`
* Create a CMake library for `lzo`
* Create a large library target for distcc
* Add an executable (named client, should probably rename it distcc to match current behavior) that links to the distcc library
* CMake generates the `config.h` header from a new `config.h.in` file using a new function called `create_config`
* Future: Add compile time options for the optional features
* Future: Compile the optional add-ons
* Future: Add the monitor support
* Future: Add gtest integration

# Request

Before I spend any more time on CMake, I'd like to know if anyone finds this useful. 

Here's some motivating questions:
* What is the minimum requested functionality for the maintainers to be comfortable merging it in? 
* Would a transition, or at least parallel support of CMake help the maintainers out?
* Would the use of CMake and [corrosion-rs](https://github.com/corrosion-rs/corrosion) make the proposed Rust migration easier?
* Would having a modern build system generator like CMake make it easier for contributors to add unit tests and thus promote maintenance of distcc?
* Would CMake allow easier maintenance of  special or little used platforms easier than autotools?


sourcefrog:
Thanks! I think moving to CMake is a good idea in general, because autoconf/etc is not so easy to deal with. I have not yet looked at the patch closely.

It would be important that merging it should not degrade test coverage. The way the tests run today is pretty convoluted, but perhaps this could get it on track to be simpler.

Ryanf55:
> Thanks! I think moving to CMake is a good idea in general, because autoconf/etc is not so easy to deal with. I have not yet looked at the patch closely.
> 
> It would be important that merging it should not degrade test coverage. The way the tests run today is pretty convoluted, but perhaps this could get it on track to be simpler.

Excellent, good to hear.



I agree, I'm happy to work with you to make the tests easier to run. CMake a nice CLI for running tests. It also has JUnit output format to allow test report parsing and display in Github. Right now, there are no negative impacts of the merge other than slightly longer clone time of the repo. This can be done in a purely additive manner, without breaking autconf or existing tests.

Ryanf55:
Per the latest commits, the following executables all now run:
* distcc 
* distccmon-gnome 
* distccmon-text

You can enable the GTK apps like so at configure time.
`cmake -B build -DWITH_GTK=ON`
I used PkgConfig to locate GTK3.

Avahi/zeroconf has been added in. It can be disabled like so:
`cmake -B build -DWITH_AVAHI=OFF`
Avahi is found with a custom find module. I did not find a PkgConfig file.

I haven't gotten to auth yet, and there's a number of TODO's, but I don't consider any of those blockers to merge the initial support.

Let me know what you think!

sourcefrog:
Thanks a lot for working on this and sorry for the delay.

I think getting away from autoconf etc would be a big improvement for devex.

I've found CMake is can also be pretty quirky and annoying. I'd like to at least consider using Meson instead because it seems potentially cleaner, but I haven't used it beyond toy projects.

I'll defer to @akarle on the direction for this. Probably either one would be an improvement.

Issue #129, opened 2015-06-06
distccd --no-detach outlives its parent process


GoogleCodeExporter:
```
trunk r780 on Debian sid

Originally reported at <http://bugs.debian.org/709169>.


NoDetachDaemon_Case leaves the daemon process running.  Maybe specific to 
Debians sh (dash 0.5.7).  I shall patch this in trunk if someone confirms this 
effects other systems.


$ PATH="`pwd`:/usr/local/bin:/bin:/usr/bin" PYTHONPATH="`pwd`/test" python2.7 
Python 2.7.3 (default, Mar  5 2013, 01:19:40) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import testdistcc
>>> obj = testdistcc.NoDetachDaemon_Case()
>>> obj.setup()
>>> print obj.pid # the process reaped by killDaemon
3905
>>> print open(obj.daemon_pidfile, "r").read()
3906

>>> 

[1]+  Stopped                 PATH="`pwd`:/usr/local/bin:/bin:/usr/bin" 
PYTHONPATH="`pwd`/test" python2.7
$ ps xf | grep distccd | grep -v grep
 3905 pts/1    T      0:00  |   \_ /bin/sh -c distccd «ARGS»
 3906 pts/1    SN     0:00  |       \_ distccd «ARGS»
 3907 pts/1    SN     0:00  |           \_ distccd «ARGS»
 3908 pts/1    SN     0:00  |           \_ distccd «ARGS»
 3909 pts/1    SN     0:00  |           \_ distccd «ARGS»
$ fg
PATH="`pwd`:/usr/local/bin:/bin:/usr/bin" PYTHONPATH="`pwd`/test" python2.7

>>> obj.killDaemon()
>>> 
$ ps xf | grep distccd | grep -v grep
 3906 pts/1    SN     0:00 distccd «ARGS»
 3907 pts/1    SN     0:00  \_ distccd «ARGS»
 3908 pts/1    SN     0:00  \_ distccd «ARGS»
 3909 pts/1    SN     0:00  \_ distccd «ARGS»
$ 


The log file confirms “not detaching”.

You can see that ‘killDaemon’ reaps the shell process and the daemon 
process survives.  To me that seems strange and perhaps a bug; is there some 
reason why the daemon should outlive the shell when ‘--no-detach’ is used?

```

Original issue reported on code.google.com by `mand...@gmail.com` on 25 May 2013 at 4:51

Issue #522, opened 2024-05-31
NoDetachDaemon_Case fails on Linux


sourcefrog:
This seems reproducible for me here on Pop_OS, although I think it may not show up in CI

```
NoDetachDaemon_Case            FAIL
-----------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mbp/src/distcc-riirbu/test/comfychair.py", line 349, in runtest
    obj.runtest()
  File "/home/mbp/src/distcc-riirbu/./test/testdistcc.py", line 885, in runtest
    self.compile()
  File "/home/mbp/src/distcc-riirbu/./test/testdistcc.py", line 910, in compile
    out, err = self.runcmd(cmd)
  File "/home/mbp/src/distcc-riirbu/test/comfychair.py", line 212, in runcmd
    raise AssertionError("""command returned %d; expected %s: \"%s\"
AssertionError: command returned 116; expected 0: "DISTCC_FALLBACK=0 distcc /bin/gcc -o testtmp.o  -c testtmp.c"
stdout:

stderr:

test_log:
Run command: /bin/cc --version
Wait status: 0x0 (exit code 0, signal 0)
stdout:
cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.


stderr:
Run command: /bin/cc --version
Wait status: 0x0 (exit code 0, signal 0)
stdout:
cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.


stderr:
Run in background:
"distccd --no-detach --daemon --verbose --log-file '/home/mbp/src/distcc-riirbu/_testtmp/NoDetachDaemon_Case/distccd.log' --pid-file '/home/mbp/src/distcc-riirbu/_testtmp/NoDetachDaemon_Case/daemonpid.tmp' --port 42000 --allow 127.0.0.1 --enable-tcp-insecure"
pid: 288561
Run command: DISTCC_FALLBACK=0 distcc /bin/gcc -o testtmp.o  -c testtmp.c
[...]

PR #548, opened 2025-07-15
tests: Fix NoDetachDaemon_Case leaving zombies around

akarle:
This test uses a helper `runcmd_background` which launches distccd with --no-detach via a shell process (`sh -c`). It then saves off the *sh* pid and kills that. Given that the distccd process resets its process group, the SIGTERM only goes to sh, and distccd becomes a child of init (continuing to run indefinitely).

This causes even greater confusion because subsequent tests "succeed" by compiling using the zombies from past runs (even though the distccd.log files show that the current distccd is failing to start).

To make matters even more confusing, this failure is shell dependent. bash treats `-c 'foo'` as "exec foo" (so the pid returned is the pid of foo) while dash treats it as "fork and exec", so the pid returned is the pid of dash.

```
$ bash -c 'sleep 1' & pstree -p $$
[1] 20191
bash(20139)─┬─pstree(20192)
            └─sleep(20191)

$ dash -c 'sleep 1' & pstree -p $$
[1] 20193
bash(20139)─┬─dash(20193)───sleep(20195)
            └─pstree(20194)
```

We noticed this in our CI, which was hanging waiting for all processes to exit.

Fixes: distcc/distcc#129

## Testing done

- `make check` passes on Debian 12

Note: this possibly is related to #522 but I can't reproduce that failure to confirm

### DIFF
--- a/test/testdistcc.py
+++ b/test/testdistcc.py
@@ -1578,7 +1578,12 @@ class NoDetachDaemon_Case(CompileHello_Case):
             time.sleep(0.2)
 
     def killDaemon(self):
-        os.kill(self.pid, signal.SIGTERM)
+        # Terminate the process specified by the pidfile.  That should kill
+        # the distccd process, any child distccd processes and the shell
+        # process used to launch distccd.
+        daemon_pid = int(open(self.daemon_pidfile, 'rt').read())
+        os.kill(daemon_pid, signal.SIGTERM)
+
         pid, ret = os.wait()
         self.assert_equal(self.pid, pid)
 


akarle:
Thanks for all the reviews!

Hm very perplexing; I've seen this error before when a previous test case fails to kill it's daemon; the NoDetach test case will fail to start distccd (port in use) and then the test will reuse the existing daemon, only to fail during cleanup because the pid file doesn't exist.

I'll investigate early this week (don't have a mac at home to repro). Sorry for the breakage!

akarle:
OK I have a minimal reproduction / good feel for why NoDetach is failing; it is as I said above--a previous test case is leaving around distccd's on the default port, which prevents NoDetach from starting up / the pidfile never exists.

The test that fails to clean up most often for me is `CPlusPlus_SystemIncludeDirectories_Case`, which is SKIPPED/NOTRUN; there's a race condition where if the SIGTERM is received during the spawning of children workers, it fails to call the signal handler and clean up. This happens often on skipped tests because they shut down the server right after starting it (also why it's sporadic in CI).

I can get the test to fail to clean up every time by setting a high `-j` number in the daemon_cmd in WithDaemon_Case.

What's bizarre is that this is macOS _only_; linux has no such problems with cleanup.

I can reproduce in a standalone fashion by starting the daemon with a very high -j and then killing it after a small sleep:

```
$ ./distccd -j1000 --no-detach --port 8080 --verbose --log-file=/tmp/distccd.log --allow=172.21.82.52 --enable-tcp-insecure & sleep 0.2 && kill $!
```

and the log stops without any cleanup:

```
distccd[90156] (dcc_create_kids) up to 189 children
distccd[90156] (dcc_create_kids) up to 190 children
[end of log]
```

Interestingly, the controlling terminal shows:
```
Trace/BPT trap: 5
```
Instead of "terminated", which appears to be the root cause (the program crashing before handling the signal).

I need to poke around some more, but I have a high degree of confidence this wasn't a regression in this PR (or any of the others we merged) but rather exposing some existing bug. I checked out v3.4 + 776515a and it repros there too.

akarle:
pulling this out into #563 to avoid burying the discussion in a merged PR! I'll look into a solution / have some ideas on how to fix it without losing too much info

Issue #515, opened 2024-04-29
Errors compiling with u8 literals


mrmbernardi:
I'm trying to compile this file: https://github.com/OpenRCT2/OpenRCT2/blob/5470a49382a5633d57d57be98b8fa90228f53375/src/openrct2/localisation/Currency.cpp

For whatever reason, when using pump mode and distcc, I get the following output:

```
./OpenRCT2/src/openrct2/localisation/Currency.cpp:25:51: error: initializer-string for char array is too long, array size is 8 but initializer has size 9 (including the null terminating character)
   25 |     {   "NLG",  10,     CurrencyAffix::Prefix,    u8"Æ<U+0092>Â ",     CurrencyAffix::Prefix,    "fl.",  STR_GUILDERS        },  // Dutch Guilder
      |                                                   ^~~~~~~~~~~~~~~
./OpenRCT2/src/openrct2/localisation/Currency.cpp:30:51: error: initializer-string for char array is too long, array size is 8 but initializer has size 10 (including the null terminating character)
   30 |     {   "CZK",  100,    CurrencyAffix::Suffix,    u8"Â KÄ<U+008D>",    CurrencyAffix::Suffix,   " Kc",  STR_CZECH_KORUNA    },  // Czech koruna
      |                                                   ^~~~~~~~~~~~~~~~
2 errors generated.
```

Perhaps there is something going on with character encoding when sending the files to the server?

akarle:
I think #549 should fix it--if you can try that patch and confirm, that'd be super helpful! (See also #481)

Issue #481, opened 2023-05-17
support utf8?


Ron2014:
utf8 program prints a mess like this
```bash
[2023-05-17 11:58:20]root@localhost(192.168.132.168):/e/GitHub/issue-test/c++/test81# export DISTCC_POTENTIAL_HOSTS='localhost 192.168.132.188'
[2023-05-17 11:58:40]root@localhost(192.168.132.168):/e/GitHub/issue-test/c++/test81# make clean && pump make -j16 CXX='distcc g++'
rm -f main main.o part.o
__________Using distcc-pump from /usr/local/bin
__________Found 1 available distcc servers
distcc g++ -std=c++11   -c -o main.o main.cc
distcc g++ -std=c++11   -c -o part.o part.cc
distcc g++ -o main -std=c++11 main.o part.o
__________Shutting down distcc-pump include server
[2023-05-17 11:58:48]root@localhost(192.168.132.168):/e/GitHub/issue-test/c++/test81# ./main
è¿éæ¯ä¸­æ 
```

but normal make, I can get the right result
```bash
[2023-05-17 11:58:55]root@localhost(192.168.132.168):/e/GitHub/issue-test/c++/test81# make clean && make
rm -f main main.o part.o
g++ -std=c++11   -c -o main.o main.cc
g++ -std=c++11   -c -o part.o part.cc
g++ -o main -std=c++11 main.o part.o
[2023-05-17 11:59:29]root@localhost(192.168.132.168):/e/GitHub/issue-test/c++/test81# ./main
这里是中文
```

codes in 3 files: main.cc part.cc part.h
```c++
// main.cc
#include <iostream>
#include "part.h"
using namespace std;

int main(int argc, char *argv[])
{
    foo();
    return 0;
}

// part.cc
#include "part.h"
#include "stdio.h"

int foo()
{
    const char *szTestInfo = "这里是中文";
    printf("%s\n", szTestInfo);
    return 0;
}

// part.h
extern int foo();
```

makefile
```makefile
CXXFLAGS=-std=c++11
CXX=g++
RM=rm -f

objects=main.o part.o
target=main

$(target): $(objects)
	$(CXX) -o $(target) $(CXXFLAGS) $(objects)

main.o: part.h
part.o: part.h

.PHONY: clean
clean:
	$(RM) main $(objects)
```

commands
```bash
## distcc download

https://github.com/distcc/distcc/releases/download/v3.4/distcc-3.4.tar.gz

## install

[...]

PR #549, opened 2025-07-15
include_server: Fix file compression for unicode files

akarle:
compress_file.py used to read in the target files as latin-1, which will degrade the contents of unicode characters. This behavior was added in this commit [1] to suppress the UnicodeDecodeErrors.

This causes a build failure when a unicode character is outside a string (i.e. a BOM), and causes build inconsistency when compiling strings with unicode (silently compiling "successfully" but mangling the characters).

This commit fixes the mangling of unicode characters by reading the file in as a binary file prior to compressing it.

Note that opening as utf-8 is not a solution, as files may be in ISO-8859 encoding, causing the read to fail.

Tests added to ensure that the behavior works (reading as latin-1 causes the new test case to fail).

[1]: 23d362910b1a6b2dca8a8d49e84620dce7a1345a

Fixes: distcc/distcc#481
Fixes: distcc/distcc#515

### DIFF
--- a/include_server/compress_files.py
+++ b/include_server/compress_files.py
@@ -98,7 +98,7 @@ class CompressFiles(object):
           # change its name.
           prefix = ""
         try:
-          real_file_fd = open(realpath, "r", encoding='latin-1')
+          real_file_fd = open(realpath, "rb")
         except (IOError, OSError) as why:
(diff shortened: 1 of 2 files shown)


sourcefrog (approved):
Great!

PR #561, opened 2025-10-24
distccd: Handle NULL from strsignal

akarle:
`strsignal` can return NULL on non-Linux platforms.  This patch gates the usage of strsignal in distccd's signal handler to handle this case gracefully.

Without this patch, it's possible that distccd will crash in the signal handler, which can cause it to fail to clean up children.

---

This is an older patch we had kicking around with a note that it was leaving hanging child processes on macOS; I can't reproduce the behavior anymore, and I'm pretty sure the root cause was #548 instead of strsignal, but it can't hurt to be a bit more cautious here (the man page confirms a NULL is returned when out of memory or for an unrecognized signal).


sourcefrog (commented):
otherwise lgtm, thanks!

PR #564, opened 2025-10-31
Remove signal-unsafe functions from handlers to avoid crash

### LINKED ISSUE #563: distccd crashes on macOS if signaled during forking
As discussed on #548 , macOS CI is failing sporadically, exposing a real issue with the signal handler crashing if invoked early (in all cases I've seen it's if it's invoked during dcc_create_kids).

It can be manually reproduced with the following one liner:

```
$ ./distccd -j1000 --no-detach --port 8080 --verbose --log-file=/tmp/distccd.log --allow=MYIP --enable-tcp-insecure & sleep 0.2 && kill $!
```

I've confirmed that its an issue with calling unsafe functions in the signal handler ([signal-safety(7)](https://www.man7.org/linux/man-pages/man7/signal-safety.7.html)).

I've got stack traces of it failing on strsignal (which mallocs on macOS):

```
* thread #1, stop reason = signal SIGSTOP
  * frame #0: 0x000000018de6f08c libsystem_platform.dylib`_os_unfair_lock_recursive_abort + 36
    frame #1: 0x000000018de6a1ac libsystem_platform.dylib`_os_unfair_lock_lock_slow + 336
    frame #2: 0x000000018dc6a740 libsystem_malloc.dylib`small_malloc_should_clear + 156
    frame #3: 0x000000018dc6a5b8 libsystem_malloc.dylib`szone_malloc_should_clear + 128
    frame #4: 0x000000018dc86770 libsystem_malloc.dylib`_malloc_zone_malloc + 156
    frame #5: 0x000000018dd819b0 libsystem_c.dylib`strsignal + 164
    frame #6: 0x000000010250ac68 distccd`dcc_daemon_terminate(whichsig=15) at dsignal.c:127:25 [opt]
    frame #7: 0x000000018de6c4e4 libsystem_platform.dylib`_sigtramp + 56
    frame #8: 0x000000018dd3ffac libsystem_c.dylib`fork + 56
    frame #9: 0x000000018dd3ffac libsystem_c.dylib`fork + 56
    frame #10: 0x000000010250ae84 distccd`dcc_create_kids(listen_fd=4) at prefork.c:123:20 [opt]
    frame #11: 0x000000010250ae24 distccd`dcc_preforking_parent(listen_fd=4) at prefork.c:92:13 [opt]
    frame #12: 0x000000010250a7ac distccd`dcc_standalone_server at dparent.c:159:15 [opt]
    frame #13: 0x000000010250a014 distccd`main(argc=<unavailable>, argv=<unavailable>) at daemon.c:262:15 [opt]
    frame #14: 0x000000010270d0f4 dyld`start + 520
```

As well as just `printf` (which signal-safety calls out too) from rs_log0:

```
* thread #1, stop reason = ESR_EC_BRK_AARCH64 (fault address: 0x0)
  * frame #0: 0x000000018fc26030 libsystem_platform.dylib`_os_unfair_lock_recursive_abort + 36
    frame #1: 0x000000018fc20b2c libsystem_platform.dylib`_os_unfair_lock_lock_slow + 308
    frame #2: 0x000000018fa84724 libsystem_c.dylib`localeconv_l + 68
    frame #3: 0x000000018fa819fc libsystem_c.dylib`__vfprintf + 172
    frame #4: 0x000000018fa8c158 libsystem_c.dylib`vsprintf_l + 204
    frame #5: 0x000000018fa922a8 libsystem_c.dylib`sprintf + 68
    frame #6: 0x000000010072bbfc distccd`rs_format_msg(buf="distccd", buf_len=4090, flags=6, fn="dcc_daemon_terminate", fmt="terminated by signal %d", va="\U0000000f") at trace.c:254:9 [opt]
    frame #7: 0x000000010072bf40 distccd`rs_logger_file(flags=<unavailable>, fn=<unavailable>, fmt=<unavailable>, va=<unavailable>, UNUSED_private_ptr=<unavailable>, log_fd=3) at trace.c:326:5 [opt]
    frame #8: 0x000000010072be3c distccd`rs_log0 at trace.c:226:15 [opt]
    frame #9: 0x000000010072be08 distccd`rs_log0(level=6, fn="dcc_daemon_terminate", fmt="terminated by signal %d") at trace.c:296:5 [opt]
    frame #10: 0x0000000100726c14 distccd`dcc_daemon_terminate(whichsig=15) at dsignal.c:134:9 [opt]
    frame #11: 0x000000018fc22de4 libsystem_platform.dylib`_sigtramp + 56
    frame #12: 0x000000018fa04ee8 libsystem_malloc.dylib`_malloc_fork_parent + 148
    frame #13: 0x000000018fa04ee8 libsystem_malloc.dylib`_malloc_fork_parent + 148
    frame #14: 0x000000019d7ce4f4 libSystem.B.dylib`libSystem_atfork_parent + 32
    frame #15: 0x000000018faa3454 libsystem_c.dylib`fork + 112
    frame #16: 0x0000000100726e00 distccd`dcc_create_kids(listen_fd=4) at prefork.c:123:20 [opt]
    frame #17: 0x0000000100726da0 distccd`dcc_preforking_parent(listen_fd=4) at prefork.c:92:13 [opt]
    frame #18: 0x0000000100726768 distccd`dcc_standalone_server at dparent.c:159:15 [opt]
    frame #19: 0x0000000100725f18 distccd`main(argc=<unavailable>, argv=<unavailable>) at daemon.c:262:15 [opt]
    frame #20: 0x000000018f86c274 dyld`start + 2840
```

If I delete all `rs_log_info` calls it avoids the crash, but I'll look into a solution that has a signal-safe logger so that we can at the very least log "terminated by signal 15".

akarle:
As discussed in #563, we're seeing the distccd daemon routinely crashing in CI due to calling signal unsafe functions in signal handlers.

I've reproduced the crash locally, seeing it crash on both `strsignal(3)` as well as `rs_log_info` (due to its use of `sprintf(3)`).

It appears there was already a calculated choice to run the risk of tracing in the handler, per the comment in cleanup.c:

     * If from_signal_handler (a boolean) is non-zero, it means that
     * we're being called from a signal handler, so we need to be
     * careful not to call malloc() or free() or any other functions
     * that might not be async-signal-safe.
     * (We do call the tracing functions, which is perhaps unsafe
     * because they call sprintf() if DISCC_VERBOSE=1 is enabled.
     * But in that case it is probably worth the very small risk
     * of a crash to get the full tracing output.)

It's unclear if we still want to take this risk (I've never witnessed a problem on Linux) or what the breakdown of macOS crashes due to strsignal vs sprintf might be, but the easiest way to move the needle for now without losing logging entirely is to disable strsignal on all platforms in signal handlers.

---

**UPDATE:** sprintf was a problem too, so I've updated the PR to remove all unsafe functions and do a best-effor logging of just the bare minimum (the signal number that terminated the program) to stderr.

Fixes #563

### DIFF
--- a/src/cleanup.c
+++ b/src/cleanup.c
@@ -71,11 +71,7 @@ void dcc_cleanup_tempfiles_from_signal_handler(void)
  * If from_signal_handler (a boolean) is non-zero, it means that
  * we're being called from a signal handler, so we need to be
  * careful not to call malloc() or free() or any other functions
- * that might not be async-signal-safe.
- * (We do call the tracing functions, which is perhaps unsafe
- * because they call sprintf() if DISCC_VERBOSE=1 is enabled.
- * But in that case it is probably worth the very small risk
- * of a crash to get the full tracing output.)
+ * (including our tracing) that might not be async-signal-safe.
  *
  * If $DISTCC_SAVE_TEMPS is set to "1", then files are not actually
  * deleted, which can be good for debugging.  However, we still need
@@ -94,7 +90,8 @@ static void dcc_cleanup_tempfiles_inner(int from_signal_handler)
      /* tempus fugit */
     for (i = n_cleanups - 1; i >= 0; i--) {
         if (save) {
-            rs_trace("skip cleanup of %s", cleanups[i]);
+            if (!from_signal_handler)
+                rs_trace("skip cleanup of %s", cleanups[i]);
         } else {
             /* Try removing it as a directory first, and
              * if that fails, try removing is as a file.
@@ -102,7 +99,8 @@ static void dcc_cleanup_tempfiles_inner(int from_signal_handler)
              * if both fail. */
             if ((rmdir(cleanups[i]) == -1) &&
                 (unlink(cleanups[i]) == -1) &&
-                (errno != ENOENT)) {
+                (errno != ENOENT) &&
+                !from_signal_handler) {
                 rs_log_notice("cleanup %s failed: %s", cleanups[i],
                               strerror(errno));
             }
@@ -119,7 +117,8 @@ static void dcc_cleanup_tempfiles_inner(int from_signal_handler)
         cleanups[i] = NULL;
     }
 
-    rs_trace("deleted %d temporary files", done);
+    if (!from_signal_handler)
+        rs_trace("deleted %d temporary files", done);
 }
 
 
--- a/src/distcc.c
+++ b/src/distcc.c
@@ -131,11 +131,7 @@ static void dcc_client_signalled (int whichsig)
 {
     signal(whichsig, SIG_DFL);
 
-#ifdef HAVE_STRSIGNAL
-    rs_log_info("%s", strsignal(whichsig));
-#else
-    rs_log_info("terminated by signal %d", whichsig);
-#endif
+    dcc_log_signal_termination(whichsig);
 
     dcc_cleanup_tempfiles_from_signal_handler();
 
--- a/src/dsignal.c
+++ b/src/dsignal.c
@@ -120,19 +120,8 @@ static void dcc_daemon_terminate(int whichsig)
 
     am_parent = getpid() == dcc_master_pid;
 
-    /* syslog is not safe from a signal handler */
-    if (am_parent && !rs_trace_syslog) {
-#ifdef HAVE_STRSIGNAL
-        char *signame = strsignal(whichsig);
-        /* on macOS, strsignal can return NULL */
-        if (signame != NULL) {
-            rs_log_info("terminated by signal %s", signame);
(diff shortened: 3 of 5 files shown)


akarle:
> Yes, maybe either a simple fprintf(stderr) or if that's not enough just write a static string?

yep! just updated; will test more extensively on macOS tomorrow, but feeling rather confident about this iteration having gotten all the cases

I considered a global `signal_log_fd` or some other metadata on the
loggers so that we could print the message to a log file if we had one
open, but the resulting metadata tracking proved somewhat ugly; stderr
is better than nothing (also consider: when using syslog, we previously
logged nothing for distccd).

sourcefrog (approved):

PR #567, opened 2025-11-03
pump: Optimize force creation of missing dirs

akarle:
C preprocessors, when processing an include like so:

    #include "foo/../bar/baz.h"

Require foo to exist; In pump-mode, if there was nothing under foo/ to send over, we need to ensure it's still created, so the include_server has a concept of forcing directories to be created by sending over a file with a special name (`forcing_technique_271828`) per directory we've seen.

Prior to this patch, the include_server would create foo/forcing_technique_271828 and then tell the distcc client to send it over (just as any other file).

This is known inefficient (hence `--no_force_dirs`), because it can cause many many `open()` calls in the include_server (one per directory that must exist).

To speed up the include_server (which is the bottleneck in our environment), this patch has the include_server tell the client it needs the files _without creating them_, and the distcc client now has special cased logic to detect these files and tell the distccd server to create them (via the normal `FILE` command, with size 0).

---

@sourcefrog -- this should be the last of our MathWorks patches for a while (there were a couple I decided didn't stand the test of time that I may need to revisit later). It's been battle tested for several years (minus the fix in `include_server_if.c`, which I caught when polishing it up, but is out of the critical path, being just used by `--scan-includes`).

Here's a scan-includes sanity check:

```
$ cat a/a.c
#include "a.h"     // in /tmp/ak/include
#include "link.h"  // in /tmp/ak/a links to /tmp/link/link.h

int a() {
    return A;
}

$ DISTCC_HOSTS=$(hostname),lzo,cpp pump distcc --scan-includes -c a/a.c -Iinclude
:$ DISTCC_HOSTS=$(hostname),lzo,cpp pump distcc --scan-includes -c a/a.c -Iinclude
__________Using distcc-pump from /local-ssd/akarle/distcc
__________Using 1 distcc server in pump mode
FILE      /tmp/ak/a/a.c
FILE      /tmp/link/link.h
FILE      /tmp/ak/include/a.h
SYSTEMDIR /usr/lib/gcc/x86_64-linux-gnu/12/include
SYSTEMDIR /usr/local/include
SYSTEMDIR /usr/include
SYMLINK   /tmp/ak/a/link.h
DIRECTORY /tmp/ak/a
DIRECTORY /tmp/ak/include
DIRECTORY /tmp/link
__________Shutting down distcc-pump include server
```

The DIRECTORY output is the successful forcing logic at work

Running a compilation shows they're sent over correctly to force the creation server side:

```
$ DISTCC_VERBOSE=1 DISTCC_HOSTS=$(hostname),lzo,cpp pump distcc -c a/a.c -Iinclude
...
distcc[1194791] (dcc_x_token_int) send NAME00000022
distcc[1194791] (dcc_x_token_string) send string '/tmp/link/forcing_technique_271828'
distcc[1194791] (dcc_x_token_int) send FILE00000000
...
```

### DIFF
--- a/test/testdistcc.py
+++ b/test/testdistcc.py
@@ -1754,6 +1754,28 @@ class ScanIncludes_Case(CompileHello_Case):
           self.assert_equal(out, '')
           self.assert_equal(err, '')
 
+class ForceDirectory_Case(CompileHello_Case):
+    """
+    Test that the forcing_technique properly creates directories even if
+    no headers are used in them (i.e. #include foo/../bar.h creates foo)
+
+    Note that its sufficient to assert that the compile succeeds under
+    pump-mode; if the technique wasn't working, it would be a compilation error
+    """
+
+    def createSource(self):
+      CompileHello_Case.createSource(self)
+      self.runcmd("mv testhdr.h test_header.h")
+      self.runcmd("ln -s test_header.h testhdr.h")
+      self.runcmd("mkdir test_subdir")
+      self.runcmd("touch test_another_header.h")
+
+    def headerSource(self):
+        return """
+#define HELLO_WORLD "hello world"
+#include "test_subdir/../test_another_header.h"
+"""
+
 class AbsSourceFilename_Case(CompileHello_Case):
     """Test remote compilation of files with absolute names."""
 
@@ -2321,6 +2343,8 @@ tests = [
          DashD_Case,
          EmptyDefine_Case,
          DashWpMD_Case,
+         ScanIncludes_Case,
+         ForceDirectory_Case,
          BinFalse_Case,
          BinTrue_Case,
          VersionOption_Case,


src/clirpc.c
```diff
@@ -266,26 +266,36 @@ int dcc_x_many_files(int ofd,
...
-        }
-
-        if (is_link) {
-            if ((ret = dcc_read_link(fname, link_points_to)) ||
-                (ret = dcc_x_token_string(ofd, "NAME", original_fname)) ||
-                (ret = dcc_x_token_string(ofd, "LINK", link_points_to))) {
-                    return ret;
-            }
-        } else {
+        if (str_endswith("/forcing_technique_271828", original_fname)) {
```
akarle: whitespace off shows this is just a new if branch; else is unchanged

sourcefrog (dismissed):
It'd be nice to have unit tests for this kind of thing, but I can understand how starting from the place we are today that might be laborious to add. So, I won't block on it.

PR #568, opened 2025-11-11
Remove references to defunct Google Code 

jchulce:
Google Code project hosting is defunct since several years ago
Update links to point to github and email contacts to the current list
Remove the old Google Code upload script as it no longer has a use


packaging/RedHat/rpm.spec
```diff
@@ -14,10 +14,10 @@ Summary: Client side program for distributed C/C++ compilations.
 Version: %VERSION
 Release: %rel
 Group: Development/Languages
-Url: https://code.google.com/p/distcc
+Url: https://github.com/distcc/distcc/
 License: GPL
-Packager: Google Inc. <opensource@google.com>
```
akarle: this one is probably OK to update though, given that the distcc project will be the packager of any future rpm's (I'm not sure this is actually used... but can't hurt to update).

akarle (changes requested):
Thank you so much for taking the time to update these!

Overall lgtm, but I think we should retain any/all old `__author__` variables--I am not a lawyer, but IIUC, Google still retains the copyright to all their contributions. I'd imagine the email is still valid too.

The URL bits look OK to me though

PR #571, opened 2026-01-29
Contrib: Ubuntu build script + ignore distccmon-* (text and gnome)

hippyau:
A Ubuntu/Debian script to install dependencies and build, optionally with --gtk monitor and also optionally --install
Added to .gitignore to ignore built gtk monitor also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 /*.exe
 distcc
 distccd
-distccmon-text
+distccmon-*
 lsdistcc
 pump
 update-distcc-symlinks
@@ -32,3 +32,4 @@ __pycache__/
 distcc-*.tar.*
 distcc-*.NEWS
 *~
+
--- a/contrib/ubuntu/build.sh
+++ b/contrib/ubuntu/build.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+# Build script for distcc on ubuntu which installs dependencies, builds, and optionally installs.
+# Just installs dependencies, builds, and shows where the binaries are.
+# If you use --install, it will also install system-wide.
+# I made it on/for ubuntu 22.04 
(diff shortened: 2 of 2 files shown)


akarle:
Hi there; thanks for the interest in helping the distcc project!

I'm not a fully qualified maintainer, so take my opinion with a grain of salt, but my personal preference would be to keep this script as personal / out of the repository. I appreciate the effort put in, but I worry that it will quickly diverge from the INSTALL instructions / become more of a maintenance hassle than an automation win. It already looks slightly incorrect in that it's missing the update-distcc-symlinks step!

If you have a strong reason to automate this / think it'd be useful for others, I'd be interested to hear it, but `autogen.sh; configure; make; make check; make install` is pretty "in model" for source distributions in my experience.

For more background, there was some discussion on changing build systems (#545, #345) that would render this script obsolete if/when I have some time to dig in there.

The .gitignore is a good catch; if you want to PR just that, I'd approve :)

P.S. In case the tone doesn't come across right--I really don't mean to be discouraging, just practical!

hippyau:
All good and no stress... I just had issues with Ubuntu supplied distcc for a long time, and finally decided I would build myself to try and resolve those issues (which seems to have helped)... they renamed pump to distcc-pump and pump never worked etc. 

It wasn't as straight forward as I'd hoped, but not too difficult.  My only hope was this might save someone an hour of figuring out the dependencies for building on Ubuntu.

Cheers :+1:

akarle:
Thanks for understanding! Indeed you are onto something--the INSTALL instructions were stale since the libpopt vendoring removal (#539)

I just put up a PR to update the instructions, which should hopefully address the same pain point as your script :) I built it in an ubuntu:latest docker container so I have reasonably high confidence it's not missing anything

I'll close this for now; thanks again for reaching out!

PR #572, opened 2026-02-17
Update INSTALL instructions for libpopt dependency

akarle:
As discussed in #571, the INSTALL script was slightly out of date. This PR also cherry picks the .gitignore change to have distccmon-gnome.

I was able to do a successful build with these new dependencies in an ubuntu:latest docker container, so I have reasonably high conviction we aren't missing any others.

Thanks @hippyau for flagging these!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 distcc
 distccd
 distccmon-text
+distccmon-gnome
 lsdistcc
 pump
 update-distcc-symlinks
--- a/INSTALL
+++ b/INSTALL
@@ -14,7 +14,7 @@ QUICK SUMMARY
 
 0. Install prerequisites.  For example, on Debian or Ubuntu, you could use
 
-      sudo apt-get install gcc make python3 python3-dev python3-setuptools autoconf
+      sudo apt-get install gcc make python3 python3-dev python3-setuptools autoconf pkg-config libpopt-dev
 
 1. Build and install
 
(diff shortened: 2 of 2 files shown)


Ryanf55 (approved):
Good improvement, I ran into this on a headless install on Ubuntu 22.

Otherwise, you hit the following:
```
checking for pkg-config... no
checking for POPT... configure: error: in `/home/distcc/distcc':
configure: error: The pkg-config script could not be found or is too old.  Make sure it
is in your PATH or set the PKG_CONFIG environment variable to the full
path to pkg-config.

Alternatively, you may set the environment variables POPT_CFLAGS
and POPT_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.

To get pkg-config, see <http://www.freedesktop.org/software/pkgconfig>.
See `config.log' for more details

```

```
checking for POPT... configure: error: Package requirements (popt >= 1.7) were not met:

No package 'popt' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables POPT_CFLAGS
and POPT_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.

```

Contrary to the install script, popt is not optional (unless there is a way to disable support for it).

akarle:
thanks for the nudge on this @Ryanf55 ; I don't have permission to merge without another review (since I was the last pusher). @sourcefrog, if you have a chance, could you approve to increase discoverability of the dependency?

sourcefrog (approved):

PR #577, opened 2026-03-04
Strip -x flag when compiling preprocessed files remotely

gorogaron:
Fixes debug info bug with CMake CMP0119 (CMake 3.20+)

When CMake policy CMP0119 is enabled, CMake adds the -x flag to explicitly specify the source language (e.g., -x c++). When distcc preprocesses locally and sends .ii files to remote servers with the -x flag present, GCC ignores line directives in the preprocessed file and records the temporary filename (e.g., /tmp/distccd_XXXXX.ii) in DWARF debug info instead of the original source path.

This breaks debugger source resolution, as the debug info points to non-existent temporary files rather than the actual source files.

Solution:
Strip the -x flag (both "-x <lang>" and "-x<lang>" formats) in dcc_strip_local_args() when sending preprocessed files to remote servers. The file extension (.ii, .mi, .mii, etc.) is sufficient for GCC to identify the language, and without -x, GCC correctly honors line directives and preserves original source paths in debug info.

Changes:
- src/strip.c: Add -x to the list of flags stripped for remote compilation
- test/testdistcc.py: Add test cases for -x flag stripping

### DIFF
--- a/src/strip.c
+++ b/src/strip.c
@@ -57,6 +57,12 @@
  * passed directly to cpp.  When given to gcc they have different
  * meanings.
  *
+ * We also strip the -x flag, which specifies the source language.
+ * When compiling preprocessed files (.ii, .mi, etc.), the -x flag
+ * causes GCC to ignore line directives and record the temporary
+ * preprocessed filename in debug info instead of the original source.
+ * The file extension is sufficient for GCC to identify the language.
+ *
  * The value stored in '*out_argv' is malloc'd, but the arguments that
  * are pointed to by that array are aliased with the values pointed
  * to by 'from'.  The caller is responsible for calling free() on
@@ -94,7 +100,8 @@ int dcc_strip_local_args(char **from, char ***out_argv)
             || str_equal("-isystem", from[from_i])
             || str_equal("-iwithprefixbefore", from[from_i])
             || str_equal("-idirafter", from[from_i])
-            || str_equal("-Xpreprocessor", from[from_i])) {
+            || str_equal("-Xpreprocessor", from[from_i])
+            || str_equal("-x", from[from_i])) {
             /* skip next word, being option argument */
             if (from[from_i+1])
                 from_i++;
@@ -110,8 +117,9 @@ int dcc_strip_local_args(char **from, char ***out_argv)
                  || str_startswith("-MT", from[from_i])
                  || str_startswith("-MQ", from[from_i])
                  || str_startswith("-isystem", from[from_i])
-                 || str_startswith("-stdlib", from[from_i])) {
-            /* Something like "-DNDEBUG" or
+                 || str_startswith("-stdlib", from[from_i])
+                 || str_startswith("-x", from[from_i])) {
+            /* Something like "-DNDEBUG" or "-xc++" or
              * "-Wp,-MD,.deps/nsinstall.pp".  Just skip this word */
             ;
         }
--- a/test/testdistcc.py
+++ b/test/testdistcc.py
@@ -478,6 +478,14 @@ class StripArgs_Case(SimpleDistCC_Case):
                  # New options stripped in 0.11
                  ("cc -o nsinstall.o -c -DOSTYPE=\"Linux2.4\" -DOSARCH=\"Linux\" -DOJI -D_BSD_SOURCE -I../dist/include -I../dist/include -I/home/mbp/work/mozilla/mozilla-1.1/dist/include/nspr -I/usr/X11R6/include -fPIC -I/usr/X11R6/include -Wall -W -Wno-unused -Wpointer-arith -Wcast-align -pedantic -Wno-long-long -pthread -pipe -DDEBUG -D_DEBUG -DDEBUG_mbp -DTRACING -g -I/usr/X11R6/include -include ../config-defs.h -DMOZILLA_CLIENT -Wp,-MD,.deps/nsinstall.pp nsinstall.c",
                   "cc -o nsinstall.o -c -fPIC -Wall -W -Wno-unused -Wpointer-arith -Wcast-align -pedantic -Wno-long-long -pthread -pipe -g nsinstall.c"),
+
+                 # Test -x flag stripping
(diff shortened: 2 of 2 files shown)


jdavidpeter:
Some more info:
The bug was observed using gcc 9 and 13 on RHEL7 and RHEL8.
The proposed fix was tested by deploying a custom build of distcc and it solved the debug symbol resolution problem for both platforms and compilers.

akarle (approved):
sorry for the delay in reviewing this! thank you for the contribution!

I confirmed that the behavior does exist with -xc on a trivial c file (the debug data had the /tmp path burned in). stripping -xc seems safe to me / results in the correct debug info.

Note that I squashed the most recent 2 commits that fixed whitespace into the main commit.

akarle:
mac CI is blocked on #599

gorogaron:
Thanks for the review @akarle. As I see #599 is merged, so could this go in as well?

akarle:
I just rebased and retriggered CI; I'll need one more approval to merge because I don't quite have full merge permissions:

> New changes require approval from someone other than akarle because they were the last pusher.

@sourcefrog not urgent, but if you can give this a second pair of eyes, that'd be great! As meta commentary, I'm hitting the second-approval needed on pretty much every PR since I'm frequently the last pusher (cleaning up the PR, etc) and/or last committer to master. If you're comfortable with me wielding a bit more commit access, that's another option (although completely understand if I'm still working my way up there).

akarle:
@gorogaron , just curious, did you use AI in any way to make this patch? We're discussing the policy in #598

jdavidpeter:
Allow me to answer, as I was working on this with @gorogaron.

The problem identification and the root-cause analysis that led us to discover the issue with the -x flag were done without AI.

A coding agent using some frontier models available at the time was then prompted to propose source code fixes for distcc. One of these proposals was selected after human review, tested manually, and then submitted as the current PR.

AI made the entire cycle much faster, especially when navigating the distcc codebase, but human intervention was needed to steer it away from some low-quality solutions.

akarle:
thanks for the transparency @jdavidpeter ; I had a hunch based on the verbose commit description

@sourcefrog I'll defer to you on if/when we want to merge this. I don't think we fleshed out whether we wanted to cut a release before accepting any (known) AI contributions

akarle (commented):
Having slept on it (and with a newfound critical eye given the AI involvement), I'm having second thoughts about this change.

This seems like possibly the wrong place to be fixing the problem. I can confirm the problem _exists_ (which is that the DW_AT_name is the /tmp file if -xc exists), but stripping the -xlang flag doesn't feel always safe--what if you have a file without an extension such that you _need_ the -x flag to tell the compiler what language it is? Would it now fail with this change if -x is stripped?

I don't have time this weekend to poke around and create an actual counter example, but wanted to flag early to discuss/put my approval on hold.

jdavidpeter:
I'd like to note again that the identification and attribution of the problem, as well as the proposed solution, are the result of human judgement. Labeling this contribution as "AI" because an LLM was used find strip.c in the codebase feels somewhat dismissive of the human engineers who worked on it.

If @akarle questions the quality of the solution - which would be normal for a PR review - he is questioning my judgement, not "AI".

Setting that aside, there are two high-level ideas behind the proposed fix:

* I could have investigated why gcc breaks the already embedded debug symbols in the preprocessed file when using -x, as this appears to be a compiler issue.
However, based on the original contract that "distcc should always generate the same results as a local compile", I felt that distcc should instead be adapted to this particular gcc behaviour.
Updating distcc to work around the issue also seems more practical than onboarding a new gcc vesion with different -x behaviour.
* distcc is a compiler wrapper, and preprocessing is an implementation detail of the compilation process. Therefore, I'm assuming that users of distcc can't directly influence the preprocessor invocation, except through the options exposed by gcc, like macro definitions.

I think the second point addresses @akarle's concern:
The original source file may indeed not have an extension, in which case -x is required to identify whether it is C or C++.
However, when preprocessing is performed locally (in non-pump mode), the preprocessor generates a temporary file using the default extension corresponding to the deduced source language.
As a result, gcc can always infer the original source type from the extension of the preprocessed output, so propagating -x when compiling an already preprocessed file is redundant.

This reasoning could break down if:

* gcc provides a way to change the name or extension of the generated preprocessed file (I'm not aware of one).
* distcc performs some internal transformation with a similar effect (I haven't checked this, but I assume it does not).
* The PR overlooks a case where it also strips -x when compiling an input that has not yet been preprocessed.

In such case, the problem may require a different approach.

Disclaimer: this response was hand crafted, then fed to ChatGPT to correct any English grammatical errors, then proofread word by word and manually stripped from AI annoyances, like em dashes or anything that I don't feel authentic.

djdomi:
> I'd like to note again that the identification and attribution of the problem, as well as the proposed solution, are the result of human judgement. Labeling this contribution as "AI" because an LLM was used find strip.c in the codebase feels somewhat dismissive of the human engineers who worked on it.

i already had a similar experience that this project dislike AI usage, thats the reason why i forked it to distcc-ng and fixed some things for my self imho @jdavidpeter

jdavidpeter:
I'm not sure forking is the right solution, as it erodes network effects and having 2 versions of distcc project running in parallel also divides and ultimately weakens community contributions.

I understand if maintainers would like to spend their time on figuring out how to chart the AI water.

I wish though we could conclude the discussion on the merits of this actual patch as running a fork in production has its own issues, and as folks are upgrading their CMake, this will hit more and more teams.

@akarle is there any input I could provide to help you?

djdomi:
> I'm not sure forking is the right solution, as it erodes network effects and having 2 versions of distcc project running in parallel also divides and ultimately weakens community contributions.
> 
TBH, my first fork actions was, to fix any open issues which was stated here open for a while :) 
However i agree on your point, but since i already got disagreed to for a support here, i will not try to make here efforts, however i still document any changes for the upstream here regardless in the hope [this is in the documentation section already known](https://github.com/distcc/distcc/discussions/598#discussioncomment-17675642) and therefore i would be happy to support here but a got rejected and i'll gonna respect this decision.

Issue #580, opened 2026-06-30
distccd: --enable-tcp-insecure should be order-insensitive with --inetd


djdomi:
> **Transparency notice:** I am not a programmer. This contribution was developed with AI assistance. I reviewed and tested the result, but the implementation was/will (be) largely AI-generated.

## Problem

The original report was about `distccd` option-order handling for TCP include related daemon flags.

The current source tree does not contain an option named `--enable-tcp-includes`. The relevant current daemon option is `--enable-tcp-insecure`.

## What This Actually Changes

The linked pull request currently adds focused regression coverage for the option-order case. It does not change the `distccd` parser implementation itself.

The test coverage checks that `distccd_parse_options()` enables the same parser state when these two argument orders are used:

```text
distccd --enable-tcp-insecure --inetd
distccd --inetd --enable-tcp-insecure
```

The expected parser result is the same in both cases:

```text
opt_enable_tcp_insecure = true
opt_inetd_mode = true
```

If a future parser change accidentally makes option order matter, the second command could behave as if TCP include handling was not enabled even though the command line still contains `--enable-tcp-insecure`:

```text
opt_enable_tcp_insecure = false
opt_inetd_mode = true
```

This means the current work proves and protects the expected behavior in tests. It should not be treated as a full runtime parser fix unless a separate parser change is added or a failing baseline reproduction proves the current parser is already affected.

## What this PR fixes

The current linked pull request fixes a missing regression-test gap around this option-order behavior.

It makes the expected parser behavior explicit in the test suite instead of relying on assumption or manual checking.

The protected behavior is that if the command line contains both `--enable-tcp-insecure` and `--inetd`, both parser flags should be enabled regardless of which option appears first.

## What changed in code

The related pull request adds a small helper, `src/h_dopt.c`, which calls the existing `distccd_parse_options()` parser with both option orders.

`Makefile.in` builds that helper, and `test/testdistcc.py` registers `TcpInsecureOptionOrder_Case` so normal test runs can exercise the coverage.

## Why this matters for users

`distccd` is commonly started indirectly. Users often do not type the final daemon command by hand; it may be assembled by a systemd unit, an inetd configuration, a container entrypoint, a distribution package, or a local wrapper script.

Those layers can place options in different orders while still expressing the same user intent. For example, these two invocations should be equivalent:

```text
distccd --enable-tcp-insecure --inetd
distccd --inetd --enable-tcp-insecure
```

[...]

PR #587, opened 2026-06-30
distccd: honor tcp include option ordering

djdomi:

Refs #580.

## Summary

Add focused regression coverage for `distccd` option-order parsing around `--enable-tcp-insecure` and `--inetd`.

## What This Actually Changes

Before this pull request, there was no focused regression test proving that `distccd` parses `--enable-tcp-insecure` the same way when it appears before or after `--inetd`.

That matters because users often do not start `distccd` by typing the final command by hand. The command line may be assembled by a systemd unit, an inetd configuration, a container entrypoint, a distribution package, or a local wrapper script. Those layers can place valid options in different orders while still expressing the same user intent.

For example, these two invocations should be equivalent:

```text
distccd --enable-tcp-insecure --inetd
distccd --inetd --enable-tcp-insecure
```

The expected parser result is the same in both cases:

```text
opt_enable_tcp_insecure = true
opt_inetd_mode = true
```

This pull request adds an executable check for that expectation. It introduces a small test helper that calls the existing option parser with both argument orders and fails if either option is lost.

For users, this does not add a new command-line option and does not change how `distccd` is normally started. The practical change is that future parser changes should now fail the test suite if they accidentally make this option order matter again.

## What this PR fixes

This pull request fixes a missing test coverage gap around `distccd` command-line option parsing. It makes the expected option-order behavior explicit in the test suite instead of leaving it as an assumption.

The current patch does not change the runtime parser implementation itself. It should therefore be treated as regression coverage for #580 unless a separate parser change is added or a failing baseline reproduction proves that the existing parser behavior is already covered by this test.

The specific behavior being protected is simple: if the command line contains both `--enable-tcp-insecure` and `--inetd`, both parser flags should be enabled regardless of which one appears first.

## What changed in code

`src/h_dopt.c` adds a small test helper for the existing `distccd_parse_options()` parser. The helper calls the parser twice with the same options in different orders:

```text
distccd --enable-tcp-insecure --inetd
distccd --inetd --enable-tcp-insecure
```

It then checks the internal parser state after each call:

```text
opt_enable_tcp_insecure = true
opt_inetd_mode = true
```

If a future parser change accidentally drops `--enable-tcp-insecure` in one order, the helper would detect a state like this and fail:

```text
opt_enable_tcp_insecure = false
opt_inetd_mode = true
```

`Makefile.in` builds the helper, and `test/testdistcc.py` registers `TcpInsecureOptionOrder_Case` so normal test runs can exercise this coverage.

## Why this matters for users

A user, package maintainer, or cluster operator may look at a service configuration and see that `--enable-tcp-insecure` is present. If parser behavior depended on option order, the visible command line could look correct while the daemon internally behaves as if that option was not enabled.

That difference is hard to diagnose from the outside. The daemon may still start, but the requested include-related daemon behavior may not match the configuration. In a distributed build setup, that can show up later as confusing pump, include transfer, or remote preprocessing behavior where one service wrapper works and another visually similar wrapper does not.

This is especially relevant for distcc because it is commonly run through wrappers, service managers, containers, and package-provided scripts. Those layers often reorder or append arguments while preserving the intended options. Equivalent option sets should produce equivalent daemon state.

Even though this pull request is test coverage, it protects that user-facing expectation by turning it into a repeatable regression test. Maintainers get a small, direct failure if future parser work makes these two valid command lines behave differently.

## Scope boundaries

This pull request currently adds regression coverage only. It does not change normal `distccd` runtime behavior and does not include release, container, RPM packaging, or unrelated Secure Shell warning fixes.

The originally reported option name `--enable-tcp-includes` is not present in the current source tree. The current daemon option is `--enable-tcp-insecure`; this pull request does not add or rename command-line options.

## Local Scope Evidence

```text
Makefile.in
src/h_dopt.c
test/testdistcc.py
```

A direct diff check against `src/dopt.c` and `src/distccd.c` is empty. A scope check found no Docker, release workflow, RPM packaging, release-package helper, or `src/ssh.c` changes in this branch.

## Validation

```text
git diff --check
./autogen.sh
make h_dopt
./h_dopt tcp-insecure-order
make TESTNAME=TcpInsecureOptionOrder_Case single-test
make check
local leak scan
```

## Changelog

I accidentally changed this pull request while testing release and packaging work. Those unrelated changes have been reversed, and the branch has been restored to the original intended scope: focused `distccd` option-order regression coverage only.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -300,6 +300,7 @@ h_scanargs_obj = src/h_scanargs.o $(common_obj)
 h_hosts_obj = src/h_hosts.o $(common_obj)
 h_argvtostr_obj = src/h_argvtostr.o $(common_obj)
 h_strip_obj = src/h_strip.o $(common_obj) src/strip.o
+h_dopt_obj = src/h_dopt.o $(common_obj) src/dopt.o src/access.o
 h_parsemask_obj = src/h_parsemask.o $(common_obj) src/access.o
 h_sa2str_obj = src/h_sa2str.o $(common_obj) src/srvnet.o src/access.o
 h_ccvers_obj = src/h_ccvers.o $(common_obj)
@@ -322,7 +323,8 @@ SRC =	src/stats.c							\
 	src/daemon.c src/distcc.c src/dsignal.c				\
 	src/dopt.c src/dparent.c src/exec.c src/filename.c		\
 	src/h_argvtostr.c						\
-	src/h_exten.c src/h_hosts.c src/h_issource.c src/h_parsemask.c	\
+	src/h_dopt.c src/h_exten.c src/h_hosts.c src/h_issource.c	\
+	src/h_parsemask.c						\
 	src/h_sa2str.c src/h_scanargs.c src/h_strip.c			\
 	src/h_dotd.c src/h_compile.c src/h_getline.c			\
 	src/help.c src/history.c src/hosts.c src/hostfile.c		\
@@ -411,6 +413,7 @@ sbin_PROGRAMS = \
 
 check_PROGRAMS = \
 	h_argvtostr@EXEEXT@ \
+	h_dopt@EXEEXT@ \
 	h_exten@EXEEXT@ \
 	h_fix_debug_info@EXEEXT@ \
 	h_hosts@EXEEXT@ \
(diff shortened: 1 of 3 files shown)


sourcefrog (commented):
I don't understand what the substantive change in this PR is meant to be? Which hunk actually fixes the bug?

Actually #580 sounds like a real bug but it's unclear what exactly is going wrong.

djdomi:

PR #588, opened 2026-06-30
pump: stabilize include scan state updates

### LINKED ISSUE #581: pump: include scan state updates should be deterministic

## Problem

Pump include-scan state updates can be hard to reason about when include file ordering is nondeterministic or when readers can observe state while it is being rewritten.

That makes repeated pump runs harder to compare and makes test failures harder to diagnose, because two runs can differ even when the logical include set is the same.

## Expected fix

The include-server file list should be returned in deterministic order, and state-file updates should publish a complete replacement rather than exposing a partially written file.

## What This Actually Changes

The linked Pull Request sorts include-server file lists before exposing them to the scan result path and writes new state through a temporary file before renaming it into place.

## Why this matters for users

Pump mode depends on include-state tracking to decide what should be available to remote compile hosts. Stable state output makes pump behavior easier to reproduce, inspect, and test.

## Linked Pull Request

Fix is proposed in #588.

djdomi:

Fixes #581.

## Summary

Stabilize pump include-scan state updates by making include scan ordering deterministic and by writing state files through a complete temporary file before replacing the visible state file.

## What This Actually Changes

Before this pull request, include scan output could depend on the order in which include files were discovered. That made pump state updates harder to compare, harder to test, and more sensitive to filesystem or traversal ordering.

This pull request makes the relevant include file lists deterministic before they are reported or used by the scan path. It also changes state file updates so readers should see either the previous complete state file or the next complete state file, not a partially written replacement.

For users, this does not add a new command-line option. The practical change is that pump include scanning should produce more stable state output, and tools watching the state file should not observe a truncated file during updates.

## What this PR fixes

This pull request fixes two stability problems in the pump include-scan path.

First, it removes avoidable ordering drift by sorting include-server file lists before they are exposed to the scan result path. Second, it makes state updates safer by writing the next state into a hidden temporary file and using `rename()` to publish the completed file.

Together, these changes make the include scan state easier to reason about and reduce race-prone behavior around readers observing state while it is being rewritten.

## What changed in code

`src/include_server_sort.c` adds a small sorting helper for include-server file lists, and `src/include_server_if.c` uses it before returning include scan results. `src/include_server_if.h` exposes the helper where the include-server interface needs it.

`src/state.c` changes state-file writing so the new contents are written to a hidden temporary file first. Once the write is complete, the temporary file is renamed into place, which is the normal pattern for publishing complete file updates.

`src/h_includesort.c` and `src/h_state.c` add focused regression helpers. `Makefile.in`, `include_server/setup.py`, and `test/testdistcc.py` wire those helpers and tests into the build and test suite.

## Why this matters for users

Pump mode depends on the include server and state tracking to decide what should be sent to remote hosts. If that state is nondeterministic or briefly incomplete while being rewritten, failures become difficult to reproduce and diagnose.

Stable ordering makes repeated runs easier to compare, which helps both maintainers and users debug pump behavior. Atomic-style state publication also reduces the chance that monitoring or test code sees a broken intermediate state instead of a complete file.

## Scope

This pull request is limited to pump include-scan ordering and state-file update behavior. It does not include release, container, RPM packaging, or unrelated Secure Shell warning fixes.

## Scope boundaries

This Pull Request is limited to deterministic pump include-scan state updates and the tests that prove the state-file behavior. It does not change release metadata, package files, container files, compression behavior, distccd daemon startup, or unrelated pump wrapper behavior.

## Local Scope Evidence

```text
Makefile.in
include_server/setup.py
src/h_includesort.c
src/h_state.c
src/include_server_if.c
src/include_server_if.h
src/include_server_sort.c
src/state.c
test/testdistcc.py
```

A direct scope check found no Docker, release workflow, RPM packaging, release-package helper, or `src/ssh.c` changes in this branch.

## Validation

```text
git diff --check
python3.13 -m py_compile include_server/setup.py test/testdistcc.py
./autogen.sh
make h_includesort h_state
./h_includesort /tmp/z /tmp/a /tmp/m
DISTCC_DIR="$PWD/_state-test" ./h_state atomic-write
make TESTNAME=IncludeServerFileOrder_Case single-test
make TESTNAME=StateFileAtomicWrite_Case single-test
make TESTNAME=ScanIncludes_Case single-test
make include-server-maintainer-check
make check
local leak scan
```

## Changelog

I accidentally changed this pull request while testing release and packaging work. Those unrelated changes have been reversed, and the branch has been restored to the original intended scope: pump include-scan ordering and state-file update stability only.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -242,6 +242,7 @@ common_obj = src/arg.o src/argutil.o					\
 	src/safeguard.o src/snprintf.o src/timeval.o			\
 	src/dotd.o 							\
 	src/hosts.o src/hostfile.o					\
+	src/include_server_sort.o					\
 	src/implicit.o src/loadfile.o					\
 	lzo/minilzo.o                                                   \
 	@ZEROCONF_COMMON_OBJS@						\
@@ -297,6 +298,8 @@ gnome_obj = src/history.o src/mon-gnome.o				\
 h_exten_obj = src/h_exten.o $(common_obj)
 h_issource_obj = src/h_issource.o $(common_obj)
 h_scanargs_obj = src/h_scanargs.o $(common_obj)
+h_includesort_obj = src/h_includesort.o $(common_obj)
+h_state_obj = src/h_state.o $(common_obj) src/state.o src/mon.o
 h_hosts_obj = src/h_hosts.o $(common_obj)
 h_argvtostr_obj = src/h_argvtostr.o $(common_obj)
 h_strip_obj = src/h_strip.o $(common_obj) src/strip.o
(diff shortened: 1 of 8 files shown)


djdomi:

PR #590, opened 2026-06-30
ssh: preserve DISTCC_SSH options across connections

### LINKED ISSUE #583: ssh: preserve DISTCC_SSH options across connections

## Problem

`DISTCC_SSH` can include required Secure Shell options such as a custom port, identity file, wrapper, or jump-host configuration.

If distcc tokenizes the environment string in-place, a later Secure Shell connection in the same process may see a modified value and lose options that were present for the first connection.

## What This Actually Changes

The linked Pull Request tokenizes a copy of `DISTCC_SSH` instead of modifying the process-owned string returned by `getenv()`.

## Why this matters for users

Remote compiles should not behave differently between the first and later Secure Shell connections in one process. Preserving the original environment value keeps user-provided transport settings stable.

## Linked Pull Request

Fix is proposed in #590.

djdomi:

Fixes #583.

## Summary

Preserve `DISTCC_SSH` options across repeated Secure Shell connections in the same distcc process.

## What This Actually Changes

Before this pull request, the Secure Shell connection path could tokenize the string returned by `getenv("DISTCC_SSH")` directly. Tokenizing modifies the string by inserting string terminators between arguments.

That is dangerous because environment strings returned by `getenv()` are process-owned storage. If the same distcc process opens more than one Secure Shell connection, the second connection may see a modified `DISTCC_SSH` value instead of the original command line.

This pull request changes the code to tokenize a copy of `DISTCC_SSH`. The environment value remains intact for later connections.

## What this PR fixes

This fixes repeated-connection behavior for custom Secure Shell commands.

For example, a user may configure:

```text
DISTCC_SSH="ssh -p 2222 -i /path/to/key"
```

The expected behavior is that every connection in the process keeps the same command and options:

```text
ssh -p 2222 -i /path/to/key host command
ssh -p 2222 -i /path/to/key host command
```

The wrong behavior is that the first tokenization can damage the stored environment value, so a later connection may effectively lose arguments such as the port or identity file:

```text
ssh host command
```

That can make one remote compile attempt work and a later one fail or connect with different authentication or transport settings.

## What changed in code

`src/ssh.c` now copies the `DISTCC_SSH` environment value before splitting it into arguments. The tokenization still produces the same Secure Shell argument vector, but it no longer mutates the process environment string.

`src/h_ssh.c` adds a focused regression helper that exercises repeated Secure Shell connection setup in one process. `Makefile.in` builds that helper, and `test/testdistcc.py` registers the regression test.

## Why this matters for users

Many distcc deployments use `DISTCC_SSH` for non-default Secure Shell behavior: custom ports, jump hosts, identity files, wrappers, or extra transport options.

Those options are often required for the remote host to be reachable at all. If distcc silently drops them after the first connection setup, the resulting failure is confusing: the environment variable still looks correct to the user, but later connection attempts behave differently inside the process.

Keeping the original environment value intact makes repeated remote compiles consistent and keeps transport configuration under the user's control.

## Scope

This pull request is limited to preserving `DISTCC_SSH` across repeated Secure Shell connection setup and adding focused regression coverage. It does not include release, container, RPM packaging, or unrelated build-warning fixes.

## Scope boundaries

This Pull Request is limited to preserving configured Secure Shell options when distcc opens Secure Shell connections and to the tests that prove that behavior. It does not change non-Secure-Shell transport behavior, pump include-server behavior, compression behavior, release metadata, package files, or container files.

## Local Scope Evidence

```text
Makefile.in
src/h_ssh.c
src/ssh.c
test/testdistcc.py
```

A scope check found no Docker, release workflow, RPM packaging, or release-package helper changes in this branch.

## Validation

```text
git diff --check
./autogen.sh
make h_ssh
make TESTNAME=SecureShellCommandEnvironment_Case single-test
make check
local leak scan
```

## Changelog

I accidentally changed this pull request while testing release and packaging work. Those unrelated changes have been reversed, and the branch has been restored to the original intended scope: preserving `DISTCC_SSH` options across repeated Secure Shell connections only.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -298,6 +298,7 @@ h_exten_obj = src/h_exten.o $(common_obj)
 h_issource_obj = src/h_issource.o $(common_obj)
 h_scanargs_obj = src/h_scanargs.o $(common_obj)
 h_hosts_obj = src/h_hosts.o $(common_obj)
+h_ssh_obj = src/h_ssh.o $(common_obj) src/ssh.o
 h_argvtostr_obj = src/h_argvtostr.o $(common_obj)
 h_strip_obj = src/h_strip.o $(common_obj) src/strip.o
 h_parsemask_obj = src/h_parsemask.o $(common_obj) src/access.o
@@ -322,7 +323,8 @@ SRC =	src/stats.c							\
 	src/daemon.c src/distcc.c src/dsignal.c				\
 	src/dopt.c src/dparent.c src/exec.c src/filename.c		\
 	src/h_argvtostr.c						\
-	src/h_exten.c src/h_hosts.c src/h_issource.c src/h_parsemask.c	\
+	src/h_exten.c src/h_hosts.c src/h_ssh.c			\
+	src/h_issource.c src/h_parsemask.c				\
 	src/h_sa2str.c src/h_scanargs.c src/h_strip.c			\
 	src/h_dotd.c src/h_compile.c src/h_getline.c			\
 	src/help.c src/history.c src/hosts.c src/hostfile.c		\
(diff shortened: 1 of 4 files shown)


djdomi:

PR #591, opened 2026-06-30
pump: fail closed on include server stalls

### LINKED ISSUE #584: pump: include server should fail closed on stalls and skipped checks

## Problem

Some pump include-server failure paths can stall, skip silently, or look successful without proving that the include-server tests actually ran.

That is unsafe for strict validation because a skipped include-server check is not proof that the include server works.

## What This Actually Changes

The linked Pull Request adds bounded failure behavior for include-server startup and request handling, adds wall-clock request timeout coverage, makes strict include-server maintainer checks fail when no include-server tests actually ran, and tightens related pump wrapper failure paths.

## Why this matters for users

Pump users usually see these problems as builds that stop making progress. Failing closed with a clear error is easier to diagnose than an include server that hangs during startup, request handling, or shutdown.

## Linked Pull Request

Fix is proposed in #591.

djdomi:

Fixes #584.

## Summary

This Pull Request makes several pump include-server failure paths fail closed instead of hanging, skipping silently, or reporting success without proving that the include-server checks actually ran.

The common theme is bounded failure behavior. If the include server cannot start correctly, cannot finish a request in time, cannot discover the required C extension, or cannot complete its maintainer checks in strict mode, the command should return a clear failure instead of leaving users with a stalled pump session or a misleading green test result.

## What This Actually Changes

Pump include-server startup and request handling are made stricter and more bounded. A child-process startup failure before readiness is now reported back instead of letting the parent wait indefinitely. Include request processing now has a wall-clock deadline in addition to the existing CPU/user-time limit, so a request that stalls while reading input is still bounded.

The maintainer include-server check is also made stricter by default. A skipped include-server test is not treated as success in strict mode, because a skipped check does not prove that the include server works. If a non-strict skip is needed, it has to be selected deliberately instead of becoming an accidental green result.

## What this PR fixes

One failure mode is a startup stall. Expected behavior is:

```text
include-server child setup fails -> parent sees the failure -> pump exits non-zero
```

The unsafe behavior is:

```text
include-server child setup fails before readiness -> parent keeps waiting -> pump appears hung
```

A second failure mode is a request that blocks while reading data. The previous CPU-time based timeout can miss that kind of stall because a blocked read may consume little or no CPU time. The added wall-clock deadline covers the user-visible case: the request takes too long in real time and must fail closed.

A third failure mode is a misleading maintainer check. If the include-server test path skips all relevant tests, the check output can look successful even though it did not verify the include server. In strict mode, that is now treated as failure.

## What changed in code

`include_server/include_server.py` reports child setup failures during startup and starts request timing before reading request data. This is the main runtime change that prevents startup and request stalls from becoming unbounded waits.

`include_server/basics.py` and the include-server tests cover the new wall-clock timeout behavior. `include_server/include_server_test.py` covers the pre-ready startup failure path so the regression cannot quietly return.

`find_c_extension.sh`, `pump.in`, and `include_server/run.py` fail closed when extension discovery or process replacement fails. `pump.in` also bounds shutdown waiting after `SIGTERM` so cleanup does not wait forever.

## Why this matters for users

Pump users usually see these problems as a build that stops making progress, not as an obvious compiler error. That is especially hard to diagnose on remote or clustered builds, because the visible symptom can be only that pump never completes startup, shutdown, or one include request.

For maintainers, the strict test behavior matters because a skipped include-server check is not proof. It can hide the exact class of problem this Pull Request is trying to catch. Making that strict by default keeps a green result meaningful.

## Scope boundaries

This Pull Request is limited to pump include-server fail-closed behavior, timeout handling, extension discovery failure handling, and the tests that prove those paths. It does not add release metadata, package files, container files, RPM or Debian packaging changes, compression changes, or unrelated daemon behavior changes.

## Local Scope Evidence

Current branch scope is limited to these files:

```text
Makefile.in
find_c_extension.sh
include_server/basics.py
include_server/basics_test.py
include_server/c_extensions/distcc_pump_c_extensions_module.c
include_server/c_extensions_test.py
include_server/include_server.py
include_server/include_server_test.py
include_server/run.py
pump.in
```

No Dockerfile, release workflow, RPM spec, Debian packaging file, or container entrypoint file is part of this Pull Request.

## Validation

Local validation for this branch should include the include-server test suite paths touched by this Pull Request, the pump startup and shutdown paths, and the strict include-server maintainer-check behavior. A warning, skipped strict include-server check, startup failure, or timeout failure must be treated as a failed validation result.

## Changelog

I accidentally changed this Pull Request while testing release and packaging work. Those unrelated changes have been reversed, and the branch has been restored to the original intended scope: pump include-server fail-closed stall and timeout behavior only.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -54,6 +54,8 @@ icondir = $(datarootdir)/pixmaps
 desktopdir = $(datarootdir)/applications
 
 include_server_builddir = $(builddir)/_include_server
+INCLUDE_SERVER_BUILD_REQUIRED = no
+INCLUDE_SERVER_CHECKS_STRICT = yes
 
 # These must be done from here, not from autoconf, because they can
 # contain variable expansions written in Make syntax.  Ew.
@@ -563,10 +565,17 @@ distccmon-gnome@EXEEXT@: $(mon_obj) $(gnome_obj)
 
 # The include-server is a python app, so we use Python's build system.  We pass
 # the distcc version, the source location, the CPP flags (for location of the
-# includes), and the build location.
+# includes), and the build location.  Keep a configurable strict mode because
+# some packaging paths intentionally tolerate builds without include-server,
+# while maintainer checks must fail if no include-server was actually built.
 include-server:
-	if test -z "$(PYTHON)"; then	\
-	  echo "Not building $@: No suitable python found"; \
+	if test -z "$(PYTHON)" || test "$(PYTHON)" = ":"; then	\
+	  if test "$(INCLUDE_SERVER_BUILD_REQUIRED)" = "yes"; then \
+	    echo "Cannot build $@: No suitable python found" 1>&2; \
+	    exit 1; \
+	  else \
+	    echo "Not building $@: No suitable python found"; \
+	  fi; \
 	else						\
 	  mkdir -p "$(include_server_builddir)" &&      \
 	  DISTCC_VERSION="$(VERSION)"			\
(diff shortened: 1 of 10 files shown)


djdomi:

PR #592, opened 2026-06-30
distccd: prune stale compile stats before accounting

### LINKED ISSUE #585: distccd: stale stats entries can affect job accounting

## Problem

The distccd compile-statistics pruning path can mishandle stale entries when the entry being removed is the head of the statistics list.

If the old head is freed and iteration continues through invalid list state, the distccd parent can crash or account jobs incorrectly while maintaining statistics.

## What This Actually Changes

The linked Pull Request saves the next list entry before unlinking and freeing a stale head entry, then continues iteration from that saved pointer.

## Why this matters for users

The distccd parent process manages compile jobs. A crash or corrupted accounting path in parent-side statistics cleanup can look unrelated to the compile request that triggered it, making the failure hard to diagnose.

## Linked Pull Request

Fix is proposed in #592.

djdomi:

Fixes #585.

## Summary

Fix pruning of stale compile-statistics entries when the stale entry is at the head of the list.

## What This Actually Changes

Before this pull request, the compile-statistics pruning code could mishandle the first entry in the linked list. When the oldest entry was also the head entry, the code freed the current node and then continued through a pointer path that was not valid for a head removal.

This pull request saves the next entry before unlinking and freeing the stale node. After pruning, iteration continues from the saved next entry instead of relying on a previous-node pointer that does not exist for the list head.

For users, this does not add a new option or change normal distcc scheduling. It makes the statistics maintenance path safer when old compile entries are removed.

## What this PR fixes

This fixes a stale statistics pruning bug in the distccd parent process.

The protected case is:

```text
old head entry -> newer entry -> newer entry
```

When the old head entry is pruned, the expected result is:

```text
newer entry -> newer entry
```

The wrong behavior is that pruning the old head can continue through invalid list state after the head has been freed. In the worst case, the distccd parent can crash while processing statistics.

## What changed in code

`src/stats.c` now stores `curr_sd->next` before unlinking and freeing the stale entry. That saved next pointer becomes the next iterator position after pruning.

`src/h_stats.c` adds a focused regression helper for the stale-head pruning path. `Makefile.in` builds the helper, and `test/testdistcc.py` registers `Stats_Case` so the behavior is covered by the test suite.

## Why this matters for users

The distccd parent process is responsible for accepting and managing compile jobs. A crash in parent-side statistics maintenance can look unrelated to the user's current compile command because it depends on background statistics state and timing.

That makes the failure hard to diagnose: a user may see distributed builds stop, workers disappear, or job accounting behave strangely even though the compile request itself is not malformed.

Fixing the head-pruning case keeps stale statistics cleanup from destabilizing the daemon. It also gives maintainers a direct regression test for the linked-list edge case that previously was easy to miss.

## Scope

This pull request is limited to compile-statistics pruning and focused regression coverage. It does not include release, container, RPM packaging, unrelated test-harness work, or unrelated Secure Shell warning fixes.

## Scope boundaries

This Pull Request is limited to distccd compile-statistics pruning when stale entries are removed from the head of the statistics list, plus the helper and regression test for that case. It does not change release metadata, package files, container files, compression behavior, or unrelated daemon scheduling behavior.

## Local Scope Evidence

```text
Makefile.in
src/h_stats.c
src/stats.c
test/testdistcc.py
```

A scope check found no Docker, release workflow, RPM packaging, release-package helper, or `src/ssh.c` changes in this branch.

## Validation

```text
git diff --check
./autogen.sh
make h_stats
./h_stats prune-old-head
make TESTNAME=Stats_Case single-test
make check
local leak scan
```

## Changelog

I accidentally changed this pull request while testing release and packaging work. Those unrelated changes have been reversed, and the branch has been restored to the original intended scope: compile-statistics stale-head pruning only.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -300,6 +300,8 @@ h_scanargs_obj = src/h_scanargs.o $(common_obj)
 h_hosts_obj = src/h_hosts.o $(common_obj)
 h_argvtostr_obj = src/h_argvtostr.o $(common_obj)
 h_strip_obj = src/h_strip.o $(common_obj) src/strip.o
+h_stats_obj = src/h_stats.o src/h_stats_stats.o $(common_obj) src/srvnet.o \
+	      src/access.o
 h_parsemask_obj = src/h_parsemask.o $(common_obj) src/access.o
 h_sa2str_obj = src/h_sa2str.o $(common_obj) src/srvnet.o src/access.o
 h_ccvers_obj = src/h_ccvers.o $(common_obj)
@@ -324,7 +326,7 @@ SRC =	src/stats.c							\
 	src/h_argvtostr.c						\
 	src/h_exten.c src/h_hosts.c src/h_issource.c src/h_parsemask.c	\
 	src/h_sa2str.c src/h_scanargs.c src/h_strip.c			\
-	src/h_dotd.c src/h_compile.c src/h_getline.c			\
+	src/h_stats.c src/h_dotd.c src/h_compile.c src/h_getline.c	\
 	src/help.c src/history.c src/hosts.c src/hostfile.c		\
 	src/implicit.c src/io.c						\
 	src/loadfile.c src/lock.c 					\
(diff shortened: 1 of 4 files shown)


djdomi:

PR #594, opened 2026-06-30
testing: isolate no-detach daemon child waits

### LINKED ISSUE #593: Apple pump tests can fail after no-detach daemon startup leaks child status

## Problem

The no-detach daemon tests can leak or consume child-process status globally, which can make later concurrent tests fail for the wrong reason.

The observed failure pattern was that `Concurrent_Case` reported a child status that belonged to earlier no-detach daemon startup handling rather than to the compile children it started itself.

## What This Actually Changes

The linked Pull Request keeps no-detach daemon startup, readiness checks, pidfile handling, retry behavior, and child waiting isolated inside the test case that owns them.

## Why this matters for users

This is test infrastructure work, not a runtime daemon feature. Cleaner child-process ownership makes CI failures more trustworthy, especially on macOS where timing and readiness behavior exposed this path.

## Linked Pull Request

Fix is proposed in #594.

djdomi:

Fixes #593.

## Summary

This Pull Request cleans up the no-detach daemon test harness so the tests wait for the right child processes and report daemon startup failures directly.

This is test infrastructure work, not a runtime distcc behavior change. The goal is to make the existing tests reliable enough that CI failures point to the real failing condition instead of being caused by leaked child process state, stale pidfiles, reused readiness sockets, or hidden daemon startup logs.

## What This Actually Changes

`NoDetachDaemon_Case` now owns the daemon startup path more strictly. It removes stale pidfiles before retry attempts, requires the daemon pidfile after socket readiness, reports no-detach daemon logs when startup fails, and retries the specific bind-failure case that can happen when a just-started daemon loses the port race.

`Concurrent_Case` no longer depends on process-global child waiting behavior from unrelated tests. The test harness is adjusted so each test case waits for the children it started instead of letting one test accidentally consume another test's child status.

The readiness loop also uses fresh socket probes. That matters on systems where reusing one socket after an early refused connection can fail to observe a daemon that becomes ready shortly afterward.

## What this PR fixes

The expected test-harness behavior is:

```text
NoDetachDaemon_Case starts daemon -> waits for that daemon -> checks readiness -> reports that daemon's logs on failure
Concurrent_Case starts its children -> waits for those children only
```

The unsafe behavior is:

```text
one test leaves or consumes child process state globally -> another test sees the wrong state -> CI fails for the wrong reason
```

Another unsafe behavior is:

```text
readiness check sees early connection refused -> reused probe does not observe later readiness -> test fails even though daemon became ready
```

The new behavior keeps the failure closer to the real cause: daemon could not start, daemon did not become ready, or the test's own child process failed.

## What changed in code

All code changes are in `test/testdistcc.py`. The patch updates no-detach daemon startup handling, pidfile cleanup, readiness probing, retry behavior, failure logging, and child waiting in the affected test cases.

There are no source changes to `distccd`, pump, the include server, compression, package metadata, release metadata, or container files in this Pull Request.

## Why this matters for users

Users do not get a new command-line option or runtime behavior from this Pull Request. The value is that CI becomes less noisy and more trustworthy. When CI reports a failure, maintainers should be able to spend time on the actual regression instead of first proving that the test harness waited for the wrong process.

This matters for downstream users too because reliable tests reduce the chance that real regressions are missed or that useful patches are delayed by false failures.

## Scope boundaries

This Pull Request is limited to no-detach daemon test-harness isolation and readiness handling. It does not add release metadata, package files, container files, RPM or Debian packaging changes, compression changes, or runtime daemon feature changes.

## Local Scope Evidence

Current branch scope is limited to this file:

```text
test/testdistcc.py
```

No Dockerfile, release workflow, RPM spec, Debian packaging file, container entrypoint file, pump source file, include-server source file, or distccd runtime source file is part of this Pull Request.

## Validation

Local validation for this branch should include the no-detach daemon tests and concurrent child-process tests touched by `test/testdistcc.py`. A failed startup, hidden daemon log, child wait mismatch, readiness timeout, or command warning must be treated as a failed validation result.

## Changelog

I accidentally changed this Pull Request while testing release and packaging work. Those unrelated changes have been reversed, and the branch has been restored to the original intended scope: no-detach daemon test-harness isolation only.

### DIFF
--- a/test/testdistcc.py
+++ b/test/testdistcc.py
@@ -1572,6 +1572,16 @@ class NoDetachDaemon_Case(CompileHello_Case):
             return os.WEXITSTATUS(status)
         return status
 
+    def _canConnectToDaemon(self):
+        # Some platforms keep a refused connection state on a socket.  Use a
+        # fresh socket for each readiness probe so later daemon readiness is
+        # observed correctly.
+        sock = socket.socket()
+        try:
+            return sock.connect_ex(('127.0.0.1', self.server_port)) == 0
+        finally:
+            sock.close()
+
     def startDaemon(self):
         max_start_attempts = 5
         attempts = 0
@@ -1599,9 +1609,24 @@ class NoDetachDaemon_Case(CompileHello_Case):
             # same port when the daemon exits with EXIT_BIND_FAILED.
             deadline = time.time() + 30
             retry = False
-            sock = socket.socket()
-            try:
-                while sock.connect_ex(('127.0.0.1', self.server_port)) != 0:
+            while not self._canConnectToDaemon():
+                result = self._collectDaemonStartupFailure()
+                if result is not None:
+                    if result == EXIT_BIND_FAILED:
+                        self.server_port += 1
+                        retry = True
+                        break
+                    self.fail("failed to start daemon: %d" % result)
+                if time.time() > deadline:
+                    self.log("distccd log before startup timeout:\n%s" %
+                             self._readDaemonLog())
+                    self.killDaemon()
+                    self.server_port += 1
+                    retry = True
+                    break
+                time.sleep(0.2)
+            else:
+                while not os.path.exists(self.daemon_pidfile):
                     result = self._collectDaemonStartupFailure()
                     if result is not None:
                         if result == EXIT_BIND_FAILED:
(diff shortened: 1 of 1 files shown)


djdomi:

PR #586, opened 2026-06-30
pump: resolve installed include server path

djdomi:
> **Transparency notice:** I am not a programmer. This contribution was developed with AI assistance. I reviewed and tested the result, but the implementation was/will (be) largely AI-generated.

## Status

This Pull Request is closed and superseded by #597.

It should not be reviewed as the active fix for #579. The original branch accidentally received unrelated release, package, and container changes while testing, so this Pull Request no longer represented the narrow include-server path fix cleanly.

## Current replacement

#597 is the clean resubmission for #579. Its current scope is only:

```text
Makefile.in
pump.in
test/pump_include_server_path_test.py
```

## Why this body was updated

This body is updated only to make the closed Pull Request history clear and to avoid leaving stale validation or scope claims in the description. It does not reopen this Pull Request and does not ask reviewers to continue reviewing this closed branch.

Issue #392, opened 2020-07-23
distcc-pump fails to start due to incorrect include_server path


marillat:
Debian bug https://bugs.debian.org/965378

Copy/paste from the bug report above

````
The distcc-pump script fails to start properly due to not being able to
find the include_server.

More specifically, the hardcoded path to the server in
/usr/bin/distcc-pump is:

    include_server='/usr/lib/distcc-pump/include_server/include_server.py'

However, the distcc-pump package installs include_server.py to the
following location:

    /usr/lib/distcc-pump/lib/python3.7/site-packages/include_server/include_server.py

The result is the following error when starting distcc-pump:

    $ DISTCC_HOSTS="localhost,lzo,cpp" distcc-pump echo foo
    __________Using distcc-pump from /usr/bin
    __________Using 1 distcc server in pump mode
    /usr/bin/python3: can't open file '/usr/lib/distcc-pump/include_server/include_server.py': [Errno 2] No such file or directory
    __________Expected a socket at '/tmp/distcc-pump.n4LiDO/socket'
    __________Could not start distcc-pump include server

Fixing the path in /usr/lib/distcc-pump resolves the issue:

    $ sed 's#pump/include_server#pump/lib/python3.7/site-packages/include_server#' \
        /usr/bin/distcc-pump > /tmp/pump-fixed

    $ env DISTCC_LOCATION=/usr/bin DISTCC_HOSTS="localhost,lzo,cpp" \
        sh /tmp/pump-fixed echo foo
    __________Using distcc-pump from /usr/bin
    __________Using 1 distcc server in pump mode
    foo
    __________Shutting down distcc-pump include server


This has been reported already to Ubuntu here:

    https://bugs.launchpad.net/ubuntu/+source/distcc/+bug/1835493

However, I thought it would be good to report here as well.

Thank you for your time and help. If I can provide any other information
that would be of use, please let me know.
````

shawnl:
Pump mode is unmaintained.

shawnl:
Correction: the ccache maintainer seems to have some interest in the python-parts of distcc (mainly pump mode).

danieldjewell:
@shawnl FWIW, the only mention of pump mode being unmaintained that I could find is right here... The current documents site (https://distcc.github.io/) specifically indicates to use ``pump`` in the "quick 60 second intro" section as does Arch's Distcc wiki page: https://wiki.archlinux.org/title/Distcc

Not disputing the fact, but perhaps a quick documentation update would be good?

[...]

Issue #219, opened 2017-08-03
ImportError: No module named distcc_pump_c_extensions


Fungie2134:
I am trying to setup a compile cluster. I installed the 2 debs, and tried to build something. OpenCV, not working, Tried to build distcc which use make, and I get this everytime : 
`owner@node:~/distcc$ export DISTCC_POTENTIAL_HOSTS='localhost 192.168.0.53'
owner@node:~/distcc$ pump make -j8 CC=distcc
__________Using distcc-pump from /usr/local/bin
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/include_server/include_server.py", line 42, in <module>
    import distcc_pump_c_extensions
ImportError: No module named distcc_pump_c_extensions
__________Expected a socket at '/tmp/distcc-pump.7Q71SE/socket'
__________Could not start distcc-pump include server
__________Warning: distcc-pump include server failed; running unpumped
__________Found 2 available distcc servers
owner@node:~/distcc$`

TafThorne:
Please could you confirm the version of Distcc and whether you used the Debian repository via `apt` to install or if you took the packages from another source?

Fungie2134:
I took the latest version on https://code.google.com/archive/p/distcc/downloads.

TafThorne:
Thank you for confirming that.

Fungie2134:
You're welcome !
I am presently trying building from sources with Github's files.

TafThorne:
Those are probably going to match what you were using.  There was some work done on master though so someone may have repaired things.

Fungie2134:
It's done and getting exactly the same issue. What I tried (on Ubuntu Server 17.04) : 
- The latests debs,
- The included package trough apt, which seems to be up - to - date,
- Building from latest sources.
All had a little issue with /usr/local/bin where pump was unable to find distcc, a simple ln -s solved the problem.

afbjorklund:
@Fungie2134: It looks like you are trying to use python2, with the python3 extensions ?
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/include_server/include_server.py", line 42, in <module>
    import distcc_pump_c_extensions
ImportError: No module named distcc_pump_c_extensions
```

But I'm seeing similar issues as well, so maybe `find_c_extension.sh` needs updating ?
```
__________Using distcc-pump from .
__________Using 1 distcc server in pump mode
[...]

PR #597, opened 2026-07-03
pump: resolve installed include server path

### LINKED ISSUE #579: pump: installed wrapper can fail to locate include_server

## Problem

Installed `pump` or `distcc-pump` wrappers can fail before compilation starts when the wrapper points to an `include_server.py` path that no longer exists.

The user-visible failure is that pump mode starts, tries to launch the Python include server, and then exits because Python cannot open the configured include-server script path.

## What This Actually Changes

The linked Pull Request keeps the configured `include_server.py` path as the first choice. If that file is missing, it asks the configured Python interpreter for its install paths and searches those Python-versioned locations for `include_server/include_server.py`.

This changes installed wrapper path resolution only. It does not change pump protocol behavior, include scanning semantics, compression, or C-extension discovery/loading.

## Expected fix

The installed wrapper should recover from a stale generated include-server path when the include server is present under the Python installation layout used by the configured Python interpreter.

## Related context

#392 - Reports the same installed `distcc-pump` failure mode: the wrapper points to an `include_server.py` path that does not exist, while the file is installed under a Python-versioned package path. The linked Pull Request addresses that path-resolution class.

#219 - Adjacent context, not fixed by the linked Pull Request. It reports pump include-server startup failures around `distcc_pump_c_extensions` import or discovery. That can look similar to users because pump fails before compilation starts, but the linked Pull Request only fixes locating `include_server.py`; it does not change C-extension discovery or loading.

## Linked Pull Requests

#586 was the first upstream submission and is closed without merge because it accidentally included unrelated release, package, and container changes.

#597 is the clean resubmission with the intended scope only:

```text
Makefile.in
pump.in
test/pump_include_server_path_test.py
```

## Validation status

The focused local validation for #597 passed for Python 3.13 path resolution, shell syntax, Python syntax, and the local build helper path. The broader local `distcc-maintainer-check` completed but reported environment-dependent `NOTRUN` cases, so it is not counted as a fully green full-platform run.

djdomi:

Fixes #579.

## Summary

This is a clean resubmission of #586 with the scope reduced back to the installed pump wrapper include-server path fix only.

The patch makes installed `pump` wrappers recover when the configured `include_server.py` path no longer exists but the include server is installed in the Python installation paths used by the configured Python interpreter.

## Related context

#392 - Reports the same installed `distcc-pump` failure mode: the wrapper points to an `include_server.py` path that does not exist, while the file is installed under a Python-versioned package path. This Pull Request addresses that path-resolution class.

#219 - Adjacent context, not fixed by this Pull Request. It reports pump include-server startup failures around `distcc_pump_c_extensions` import or discovery. That can look similar to users because pump fails before compilation starts, but this Pull Request only fixes locating `include_server.py`; it does not change C-extension discovery or loading.

## What This Actually Changes

Before this Pull Request, an installed `pump` wrapper could rely on one rewritten `include_server.py` path. If that path was stale or did not match the active Python install layout, pump startup failed before any distributed compile work could start.

After this change, `pump.in` still keeps the configured `include_server` path as the first choice. If that file is missing, it asks the configured Python interpreter for its `sysconfig` `purelib` and `platlib` paths and searches those locations for `include_server/include_server.py`.

The fallback is tied to the Python interpreter used by the wrapper. For example, a wrapper configured for Python 3.13 searches Python 3.13 install paths instead of guessing a generic directory.

## What this PR fixes

The failure pattern is:

```text
pump --startup
python: can't open file '.../include_server.py': No such file or directory
Could not start distcc-pump include server
```

The expected behavior after this change is:

```text
pump --startup
configured include_server path is missing
pump asks Python sysconfig for install paths
pump finds include_server/include_server.py under the installed Python path
pump starts the include server
```

This does not change the pump protocol, compression, remote compile behavior, or include scanning semantics. It fixes how the installed shell wrapper locates the Python include-server entry point.

## What changed in code

`pump.in` adds helper functions to query the configured Python interpreter for its major/minor version and `sysconfig` install paths. It writes candidate paths to a temporary file and reads them line-by-line so paths containing whitespace are not broken by shell word splitting.

`Makefile.in` adds `test/pump_include_server_path_test.py` to the test sources, makes `maintainer-check-no-set-path` depend on the generated `pump` script, and runs the new regression test after the existing `test/testdistcc.py` path succeeds.

`test/pump_include_server_path_test.py` builds a fake installed layout, intentionally sets the wrapper's recorded `include_server` path to a missing file, places fake include servers in the Python-derived candidate locations, and verifies that the wrapper can still run a command through pump startup and shutdown.

## Why this matters for users

Pump mode depends on the Python include server. If the installed wrapper cannot find `include_server.py`, users see pump fail immediately even though the package may contain the include server in a valid Python install location.

This is especially visible on distributions where Python package paths include a versioned directory such as `python3.13/site-packages`. The wrapper should follow the Python interpreter it was configured with instead of depending on one stale generated path.

For users, the practical result is that an installed `pump` or `distcc-pump` command has a better chance of starting correctly after installation or Python layout changes, without changing the rest of pump behavior.

## Scope boundaries

This Pull Request is limited to installed pump wrapper include-server path resolution and the regression test for that behavior.

Current branch scope is only:

```text
Makefile.in
pump.in
test/pump_include_server_path_test.py
```

It does not include Docker files, release workflows, RPM or Debian packaging files, container entrypoints, compression changes, Secure Shell changes, daemon option parsing, or unrelated test-harness changes.

## Validation

Passed locally:

```text
git diff --check origin/master...HEAD
python3.13 --version
python3.13 -m py_compile test/pump_include_server_path_test.py
sh -n pump.in
python3.13 test/pump_include_server_path_test.py ./pump /usr/bin/python3.13
make -j1 check_programs
local leak scan on the diff and Pull Request text
```

Additional local check:

```text
make -j1 distcc-maintainer-check
```

The `distcc-maintainer-check` command completed, but this local environment reported `NOTRUN` for Objective-C, Objective-C++, system include-directory, and system-specific assembly cases. I am not counting that as a fully green full-platform run under the strict local warning/skip policy.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -698,7 +698,7 @@ TESTDISTCC_OPTS =
 # TODO(klarlund): the outermost if assumes that the include-server target may be
 # satisfied w/o actually building an include server (or rather the C extension);
 # this logic needs to be verified or amended.
-maintainer-check-no-set-path:
+maintainer-check-no-set-path: pump
 	@if ! $(PYTHON) -c 'import sys; print(sys.version)'; then \
 	  echo "WARNING: python not found; tests skipped"; \
 	else \
--- a/Makefile.in
+++ b/Makefile.in
@@ -201,6 +201,7 @@ man8dir = $(mandir)/man8
 
 test_SOURCE = test/comfychair.py \
 	test/onetest.py \
+	test/pump_include_server_path_test.py \
 	test/testdistcc.py \
 	find_c_extension.sh
 
@@ -708,7 +709,9 @@ maintainer-check-no-set-path:
 	    echo "Please change RESTRICTED_PATH to change this PATH value."; \
 	    exit 1; \
 	  fi; \
-	  $(PYTHON) "$(srcdir)/test/testdistcc.py" $(TESTDISTCC_OPTS); \
+	  $(PYTHON) "$(srcdir)/test/testdistcc.py" $(TESTDISTCC_OPTS) && \
+	  $(PYTHON) "$(srcdir)/test/pump_include_server_path_test.py" \
+	    "$(builddir)/pump" "$(PYTHON)"; \
 	fi
 
 distcc-maintainer-check: check_programs
--- a/pump.in
+++ b/pump.in
@@ -217,6 +217,56 @@ PrintIncludeServerStatusMessage() {
   fi
 }
 
+PythonMajorMinorVersion() {
+  "$PYTHON" -c 'import sys; print("%d.%d" % sys.version_info[:2])'
+}
+
+ResolveInstalledIncludeServer() {
+  if [ -n "$include_server" ] && [ -f "$include_server" ]; then
+    return 0
+  fi
+
+  python_version=`PythonMajorMinorVersion` || {
+    echo "$program_name: error: can't determine Python version from $PYTHON" \
+      1>&2
+    return 1
+  }
+
+  distcc_parent=`dirname "$distcc_location"`
+
+  # Prefer the Python version used to configure this pump script.  This keeps
+  # path resolution deterministic when multiple Python versions are installed.
+  for candidate in \
+    "$prefix/lib/python$python_version/dist-packages/include_server/include_server.py" \
+    "$prefix/lib/python$python_version/site-packages/include_server/include_server.py" \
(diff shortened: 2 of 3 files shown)


djdomi:

